PR #41, opened 2018-04-17
Simplify getting started with readme/makefile updates

pearkes:
This just makes it a bit easier for people to contribute and tries to point them in the right direction for testing.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -46,7 +46,7 @@ test-compile:
 
 test-serv: fmtcheck
 	@docker pull "consul:$(CONSUL_VERSION)"
-	docker run -p 127.0.0.1:8500:8500 "consul:$(CONSUL_VERSION)"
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -43,5 +44,9 @@ test-compile:
 	fi
 	go test -c $(TEST) $(TESTARGS)
 
-.PHONY: build test testacc vet fmt fmtcheck errcheck vendor-status test-compile
+test-serv: fmtcheck
+	@docker pull "consul:$(CONSUL_VERSION)"
+	docker run -p 127.0.0.1:8500:8500 "consul:$(CONSUL_VERSION)"
```
vancluever: Should we add `--rm` here to delete the test container after it exits?

pearkes: Yeah good call, added that. Thanks!

PR #55, opened 2018-07-17
Add Node meta to consul_node

lawliet89:


website/docs/r/node.html.markdown
```diff
@@ -30,9 +30,12 @@ The following arguments are supported:
 * `name` - (Required) The name of the node being added to, or
   referenced in the catalog.
 
+* `node_meta` - (Optional) KV metadata pairs
```
pearkes: If you can reword to:

> (Optional, map) Key/value pairs that are associated with the node.

website/docs/r/node.html.markdown
```diff
@@ -30,9 +30,12 @@ The following arguments are supported:
...
+
 ## Attributes Reference
 
 The following attributes are exported:
 
 * `address` - The address of the service.
 * `name` - The name of the service.
+* `node_meta` - (https://www.consul.io/docs/agent/http/catalog.html#Meta) KV metadata pairs
```
pearkes: I don't think this link works. I would just word this the same as the above:

> `node_meta` – Key/value pairs that are associated with the node.

consul/resource_consul_node.go
```diff
@@ -67,10 +73,20 @@ func resourceConsulNodeCreate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+	} else {
+		nodeMeta = nil
+	}
+
 	registration := &consulapi.CatalogRegistration{
 		Address:    address,
 		Datacenter: dc,
 		Node:       name,
+		NodeMeta:   nodeMeta,
```
pearkes: Just alternatively you could assign this after to `registration` instead of doing `else { nodeMeta = nil}`. Somewhat more concise and leaves the pattern open for more attributes in the future.

consul/resource_consul_node.go
```diff
@@ -34,6 +34,12 @@ func resourceConsulNode() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"node_meta": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
```
pearkes: I don't think this should be `ForceNew`. I believe node meta can be updated. This resource doesn't support updating though, currently. Thoughts?

website/docs/r/node.html.markdown
lawliet89: Fixed

website/docs/r/node.html.markdown
lawliet89: Fixed

consul/resource_consul_node.go
lawliet89: Made a change in 1f440f0. Not sure if this is what you mean.

Sorry, I am new to Go.

consul/resource_consul_node.go
lawliet89: I'm not too familiar with the internals of Terraform. As far as I can tell, there is no update mechanism to catalog registration, so every change is basically a re-create.

catsby: Registering Nodes in the catalog is an HTTP `PUT` operation, so in theory here we should be safe to remove `ForceNew` from `node_meta` (as well as `address` and `name`, I think). Both `Create` and `Update` can point to the same existing method as we have now, `resourceConsulNodeCreate`. 

This only works because the Consul API library we're using uses a `string` for the `ID` and not `*string`, meaning when we initially register the node, we send an empty string as the UUID for the node. This resource does not record that UUID and so can't re-send it as you would expect to update something with `PUT`, but since it's an empty string it matches anyway.

catsby: There is some other things going on here too, because there is no UUID being used. If `ForceNew` is removed, we need to make sure we read `node_meta` into state in the `resourceConsulNodeRead` method, to detect any change

lawliet89: I've removed the `ForceNew` and added a test for updating.

> This only works because the Consul API library we're using uses a string for the ID and not *string, meaning when we initially register the node, we send an empty string as the UUID for the node. This resource does not record that UUID and so can't re-send it as you would expect to update something with PUT, but since it's an empty string it matches anyway.

Sounds pretty ripe for future problems. Do we need to update this behaviour?

lawliet89:
@catsby and/or @pearkes could I trouble you to look through my additional changes please?

lawliet89:
Bump for reminder @pearkes. It's been > 1 month.

PR #62, opened 2018-09-10
Fix example code in consul_key_prefix doc

alkalinecoffee:
Example is missing the `subkeys` attribute.


pearkes:
Thanks! This came in #67...I think I tripped on GitHub and both of your PRs are merged with the same change, but its fixed so leaving it at that.

Issue #54, opened 2018-07-17
[Feature] Support health checks in `consul_service` resource


lawliet89:
### Affected Resource(s)
- `consul_service`

I would like to request for the ability to define health checks for services registered with the `consul_service` resource. 

As far as I understand, the resource is meant for use with [external services](https://www.consul.io/docs/guides/external.html) and it would be great if we can also define health checks for the external service. Then, we can use something else like [Consul ESM](https://github.com/hashicorp/consul-esm) to do the actual health checking.

pearkes:
I agree, this would be a good addition.

remilapeyre:
Hi @pearkes, what is the status of this proposal? I don't know go but have the same need than @lawliet89 and could try to implement it if nobody have time to spend on it.

remilapeyre:
Hi, I tried to make a PR to add this feature, I never wrote Go before so it's probably hideous. I had to make a change in `vendor/github.com/hashicorp/consul/api/catalog.go` which is probably bad. Do you want me to make a new PR in `hashicorp/consul` ?

pearkes:
@remilapeyre thanks for opening that PR!

Yes, for us to change the API interface that is vendored we need to update the upstream at hashicorp/consul (which you could make a separate PR for) and then re-vendor here with `govendor fetch github.com/hashicorp/consul@SHA`.  I can keep an eye out for that PR an d once that is done take a closer look here.

pearkes:
@remilapeyre awesome I think we're closer! Once you get that updated I'll take a closer review.

pearkes:
This is done via https://github.com/terraform-providers/terraform-provider-consul/pull/64.

PR #64, opened 2018-09-17
Add support for health-checks to consul_service

remilapeyre:
As proposed in #54, this patch add the support for health-checks defined inline in Consul services definition.


pearkes:
Made a few notes about this in the issue: https://github.com/terraform-providers/terraform-provider-consul/issues/54#issuecomment-431900444

remilapeyre:
I used this version of the code for a few weeks and I think there is some problem for the update, I need to add some tests.

pearkes:
@remilapeyre I feel like we should just get this merged in the current approach, and not necessarily wait for that upstream change. Does that sound good to you?

remilapeyre:
Hi @pearkes, I'm getting confused about the issue, if I understand https://github.com/hashicorp/consul/issues/1680 correctly, the information is already available since Consul v1.0.1.

Health-Checks definitions were first added in a7c42a6c2a3c63331170e2fee6ca1bc7bf6731e9 (Expose SkipNodeUpdate field and some health check info in the http api), then refined in subsequents commits.

Here's what http://localhost:8500/v1/health/checks/example-external gives since v1.0.1:

```
[
    {
        "Node": "external-example",
        "CheckID": "service:redis1",
        "Name": "Redis health check",
        "Status": "critical",
        "Notes": "Script based health check",
        "Output": "",
        "ServiceID": "example-external",
        "ServiceName": "example-external",
        "ServiceTags": [],
        "Definition": {
            "HTTP": "https://www.google.com",
            "Interval": "5s",
            "Timeout": "1s"
        },
        "CreateIndex": 9,
        "ModifyIndex": 10
    }
]
```

It seems to me that it is just the documentation at https://www.consul.io/api/health.html#list-checks-for-service that is not up to date.

If this is right, this PR need documentation and to make header support in checks work.

pearkes:
I'm suggesting we change Consul to return the health checks in the same API response as when querying for service instances. Currently it just returns the health check status, not definition.

But in the interim I'm supportive of this current approach of using the (separate) heath check API to read the health check definitions. Would just be nice if we could do it in one API call, but not a big deal.

> this PR need documentation and to make header support in checks work.

That makes sense to me!

consul/resource_consul_service_test.go
```diff
@@ -171,6 +205,90 @@ resource "consul_service" "example" {
...
+	node = "${consul_node.compute.name}"
+	port = 80
+
+	check {
+		check_id = "service:redis1"
+		name = "Redis health check"
+		notes = "Script based health check"
+		status = "passing"
+
+		definition {
```
pearkes: I basically think we should not have "definition" here and just allow specification under the `check {}` block, even though the API has definition on read. Does that make sense? Here's an example for how someone could write a check:


```
check {
    id = "service:redis1"
    name = "Redis health check"
    notes = "Script based health check"
    status = "passing"
    tls_skip_verify = false
    method = "PUT"
    interval = "5s"
    timeout = "1s"
    deregister_critical_service_after = "30s"
}
```

consul/resource_consul_service.go
```diff
@@ -68,6 +90,88 @@ func resourceConsulService() *schema.Resource {
...
+							Optional: true,
+						},
+
+						"status": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "critical",
+						},
+
+						"definition": {
```
pearkes: I don't think we should actually nest this. See a note below for what I mean.

consul/resource_consul_service_test.go
pearkes: This just seems cleaner and "more obvious" to me. It mirrors what you would expect I think writing check definitions in HCL or JSON.

remilapeyre: Yes, nesting doesn't add any value.

Issue #40, opened 2018-02-22
NodeMeta


Zeldhyr:
Hi,
I need the NodeMeta feature (https://www.consul.io/api/catalog.html#nodemeta).
Will it be implemented soon?
I've done a pull request (#38) for this. What do you think about it?
Thanks for your help

### Terraform Version
0.11.3

moserke:
Thanks for doing this @Zeldhyr!  We needed the same thing and I was just starting to look in to it when I saw this issue.  Hopefully this makes it into the provider soon.

pearkes:
That change in #38 looks good -- just need tests and we can get it merged!

PR #65, opened 2018-09-21
Add node-meta

pearkes:
Minor additions to the PR from @lawliet89 and @Zeldhyr renaming "node_meta" to "meta" in the TF resource. This aligns well with Terraform, and it shows up in both ways on the Consul API.

Closes #55 #38 #40


pearkes:
Thanks for the contributions here all -- tests passing and verified some of the update behavior we were concerned about.

Issue #59, opened 2018-08-21
Enhancement: Add flags support to Consul key management


delphyne:
Preamble: Consul allows you to set an unsigned integer along with a KV that can be used however the provider/consumer wish.  Often, this data contains a bitmask to let consumers know metadata about the key being read (is this value encrypted?, is it json encoded?, is the data binary?).  Terraform current does not support management of the Flags attribute.

### Terraform Version
0.11.7

### Affected Resource(s)
- consul_keys
- consul_key_prefix

### Terraform Configuration Files
proposal for interface changes...
#### consul_keys
because the `consul_keys` resource uses a sub-object, adding a flags attribute is a non-breaking change (with a caveat).  See "expected behavior" below.
```hcl

resource "consul_keys" "app" {
  datacenter = "nyc1"
  token      = "abcd"

  # Set the CNAME of our load balancer as a key
  key {
    path  = "service/app/elb_address"
    value = "${aws_elb.app.dns_name}"
    flags = 64
  }
}
```
#### consul_key_prefix
Consul key prefix uses a map to describe the subkeys, as such, a larger contract change would be necessary.  There are a couple approaches to this, but the most backwards compatible way I can think of is...
```hcl
resource "consul_key_prefix" "myapp_config" {
  datacenter = "nyc1"
  token      = "abcd"

  # Prefix to add to prepend to all of the subkey names below.
  path_prefix = "myapp/config/"

  subkeys = {
    "elb_cname"         = "${aws_elb.app.dns_name}"
    "s3_bucket_name"    = "${aws_s3_bucket.app.bucket}"
    "database/hostname" = "${aws_db_instance.app.address}"
    "database/port"     = "${aws_db_instance.app.port}"
    "database/username" = "${aws_db_instance.app.username}"
    "database/name"     = "${aws_db_instance.app.name}"
  }

  subkey {
    path="database/password"
    value="${aws_db_instance.app.password}"
    flags=64
  }
}
```
it could either be ok to provide both subkeys and [many] subkey elements, or only one or the other could be allowed on any resource.

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
Terraform should be able to manage the Flags attributes of a KV, and should ensure that the Flags of a managed KV are always correct.  (See Actual Behavior)

### Actual Behavior
[...]

PR #71, opened 2018-11-04
Add flags support to resource consul_keys

remilapeyre:
Add support for the flags parameter to the `consul_keys` resource.

#59


consul/resource_consul_keys_test.go
```diff
@@ -31,12 +32,31 @@ func TestAccConsulKeys_basic(t *testing.T) {
 					testAccCheckConsulKeysValue("consul_keys.app", "enabled", "true"),
 					testAccCheckConsulKeysValue("consul_keys.app", "set", "acceptanceUpdated"),
 					testAccCheckConsulKeysRemoved("consul_keys.app", "remove_one"),
+					testAccCheckConsulKeysFlags("test/set", 64),
```
pearkes: Do you know why we can't use the built-in helpers here to check the set? I feel like that could be used here but happy to be shown the other way.

pearkes: Yeah I think I was, but no requirement, I will approve either way (still needs a rebase though)

remilapeyre: I fixed it, thanks!

Issue #79, opened 2018-12-21
feature request: support ca path in provider settings


eedwards-sk:
this is a feature request to support the [ca_path](https://www.consul.io/docs/agent/options.html#ca_path) configuration option from consul

currently the only supported provider feature is `ca_file`, which requires a path to a single file

because a valid configuration for consul is to have a directory containing ca files used, I believe that the consul terraform provider should support it, as well

remilapeyre:
Hi @eedwards-sk, thanks for opening a feature request!

This seems reasonable and it is the last TLS configuration attribute missing for the provider.  PR #80 should do what you want.

PR #80, opened 2018-12-21
Add ca_path attribute to provider configuration

remilapeyre:
Close https://github.com/terraform-providers/terraform-provider-consul/issues/79


eedwards-sk:
Thanks for the PR for #79!

I noticed you added `CONSUL_CAPATH` env var support, as well.

I noticed that a few of the env vars supported in `resource_provider.go` do not match those supported by the [consul cli](https://www.consul.io/docs/commands/index.html#environment-variables), and some are missing entirely (e.g. `CONSUL_HTTP_SSL_VERIFY`).

Is that intentional?

Since the [provider documentation](https://www.terraform.io/docs/providers/consul/index.html#environment-variables) references the above consul cli docs page, I assume there is some intent of parity.

remilapeyre:
Thanks for the feedback!

>I noticed that a few of the env vars supported in resource_provider.go do not match those supported by the consul cli, and some are missing entirely (e.g. CONSUL_HTTP_SSL_VERIFY).

I think many environment variables are already set when instantiating the client, for example `CONSUL_HTTP_SSL_VERIFY` is used here https://github.com/terraform-providers/terraform-provider-consul/blob/1a36214955ccd62b0750154aec46849a58af3df9/vendor/github.com/hashicorp/consul/api/api.go#L354-L362 wich is called by `resource_provider.providerConfigure`.

Did you try to set `CONSUL_HTTP_SSL_VERIFY` and it did not work?

> Since the provider documentation references the above consul cli docs page, I assume there is some intent of parity.

I think so too, must would expect the behavior to be the same and it could be suprising if not for someone whose environment variables are set in `bashrc` for example. I you find some environment variables for which the behavior differ, please tell me and I will add support for them.

PR #83, opened 2019-01-07
[AUTOMATED] Upgrade to Go 1.11

appilon:


blalor:
Nit: I don't see that the README was actually updated. 🙂

appilon:
@blalor Indeed my tool picks up the version of Go from travis and then tries to find it in the Readme. My commit message is quite naive and assumed everything "worked". Fixed now great catch!

PR #85, opened 2019-01-10
Fix example HCL for consul_agent_config data resource

therealbill:
The listed HCL example does not work in terraform 11.11. The output structure is incorrect (needs value instead of a variable name) as is the reference format (missing the name section). The new example works as tested on v0.11.11 against consul 1.4.


remilapeyre:
Thanks @therealbill 👏

Issue #46, opened 2018-05-01
[ANN] Upcoming changes to Consul Provider


pearkes:
In order to be transparent with the roadmap and broadcast some substantial changes coming to the provider, I'm opening this issue in the hopes that interested users will see it. Any changes will be versioned and included in changelogs, with old versions continuing to work (as long as the upstream Consul API continues to support them). You can learn more about pinning a provider version [here](https://www.terraform.io/docs/configuration/providers.html#provider-versions).

In time I will create milestones within GitHub with some stand-alone issues for each of the below changes. Timeline across this will likely be flexible -- any help is welcome, just feel free to make an comment here.

These changes are primarily to simplify the number of resources, ensure we are using the correct APIs/design provided by Consul for something like Terraform, and remove resources that can no longer be supported by the current version of the Consul API.

## Existing Resources

Name | Type | Action | Version
-- | -- | -- | --
consul_agent_self | Datasource | Deprecate/Remove | 1.1.0/1.2.0
consul_agent_config | Datasource | Create | 1.1.0
consul_catalog_nodes | Datasource | Rename (consul_nodes) | 1.1.0
consul_catalog_services | Datasource | Rename (consul_services) | 1.1.0
consul_catalog_service | Datasource | Rename (consul_service) | 1.1.0
consul_node | Datasource | Create | 1.1.0
consul_keys | Datasource | No change |  n/a
consul_agent_service | Resource | Deprecate/Remove | 1.1.0/1.2.0
consul_catalog_entry | Resource | Deprecate/Remove | 1.1.0/1.2.0
consul_keys | Resource | No change |  n/a
consul_key_prefix | Resource | No change |  n/a
consul_node | Resource | No change |  n/a
consul_service | Resource | No change |  n/a

## New Resources

These are resources that we intend to add in the future.

Name | Type | Version
-- | -- | --
consul_autopilot_health | Datasource | 1.3.0
consul_autopilot_config | Resource | 1.3.0
consul_network_segments | Resource | 1.3.0
consul_network_area | Resource | 1.4.0
consul_network_area_members | Datasource | 1.4.0
consul_acl_token* | Datasource | 1.5.0

*\* This assumes that the Consul API supports some notion of a "token accessor", otherwise I don't think it would be wise to support it given the security implications.*

vchan2002:
Thanks for the FYI in regards to this... 

[...]

PR #86, opened 2019-01-15
Add the consul_autopilot_config resource

remilapeyre:
See https://github.com/terraform-providers/terraform-provider-consul/issues/46#issue-319347046

### DIFF
--- a/consul/resource_consul_autopilot_config.go
+++ b/consul/resource_consul_autopilot_config.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/hashicorp/errwrap"
+
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/terraform/helper/schema"
 )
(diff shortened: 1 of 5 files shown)


consul/resource_consul_autopilot_config.go
```diff
@@ -0,0 +1,150 @@
...
+			},
+			consulAutopilotConfigCleanupDeadServers: &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			consulAutopilotConfigLastContactThreshold: &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "200ms",
```
pearkes: Do you think we should just omit this and let Consul set the default upstream? That default may change in the future. It might be better to say this is computed if not set.

website/docs/r/autopilot_config.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+* `disable_upgrade_migration` - (Optional) Whether to disable [upgrade migrations](https://www.consul.io/docs/guides/autopilot.html#redundancy-zones) Defaults to false.
+
+* `upgrade_version_tag` - (Optional) The tag to override the version information used during a migration. Defaults to an empty string.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `datacenter`
```
pearkes: Although it is receptive I think it helps to copy paste the description for these attributes.

consul/resource_consul_autopilot_config.go
remilapeyre: I first tried to do this with `Operator.AutopilotCASConfiguration` but I got issues with defaults which would always differ from the state and that would result in a perpetual diff.

I did not think of it at the time but could I use `DefaultFunc` to make a call to Consul and fetch current settings here? I think it would make one call per attribute unless I find a way to cache the call. This behavior would be more predictable to users and the extra complexity is probably worth it.

pearkes: Hm I think using defaultfunc would basically just be doing `computed: true` which I think makes sense for this. Do you see what I mean? It is a computed attribute meaning when it isn't set it will be computed remotely and then written back.

remilapeyre: I tried that but it seems that setting `Computed: true` it is then impossible to detect when between:
```
resource "consul_autopilot_config" "config" {}
```
and
```
resource "consul_autopilot_config" "config" {
	upgrade_version_tag =  ""
}
```
 The first should be no-op but the second should remove the current value of `upgrade_version_tag` and set `""` instead. It seems to me that it is not possible to detect whether a value has been defined by the user when it is computed: https://github.com/hashicorp/terraform/blob/3d908f56aa6588b9a2db7cdf6b36bdf7b45d781f/helper/schema/resource_data.go#L123

Maybe a `GetOkExistsComputed` could be used here?

pearkes: Hm from the docs there it seems like that is only intended for booleans. Perhaps we should just hardcode the defaults for now. It isn't ideal but I'm comfortable with it given the constraints as I understand them.

Issue #87, opened 2019-01-17
[Feature Request] data.consul_service using health endpoint


eedwards-sk:
Currently, `data.consul_service` has almost no use for me, as I discovered it returns from the catalog and does not exercise the health check results.

Ideally, I'd like to use `data.consul_service` to return the healthy addresses for a service. I'm not sure what use cases people have where they want to get unhealthy results, but I don't have one.

I need to configure terraform's vault provider with the address of a healthy vault instance. With the current `data.consul_service` behavior, I end up getting back node addresses of unhealthy or left nodes.

remilapeyre:
Hi @eedwards-sk, can you give more information about your use-case?

It is true that a new data source may be needed to fetch health information about a given service but maybe you can use Consul DNS interface to achieve the same goal.

eedwards-sk:
Hi @remilapeyre thanks for the response

Unfortunately I cannot use dns, and I tried many, many ways to make that work first (I know more about iptables and resolv.conf now than I ever cared to).

I'm doing orchestration [running terraform as concourse-ci tasks](https://github.com/Snapkitchen/concourse-terraform) and inside containers you cannot override resolvers on alpine, for example (thanks to how it queries all dns servers regardless of order), and most docker hosts (or even in my case, runc with concourse) also usually controls the resolv.conf, so it's not always possible to redirect dns to a consul agent.

I tried to state the use case above, but I'll reiterate and try to expand on it further:

>I need to configure terraform's vault provider with the address of a healthy vault instance. With the current data.consul_service behavior, I end up getting back node addresses of unhealthy or left nodes.

from `shared.tf` on a project where I'm setting some vault configuration:

```hcl
# ==========
# data sources
# ==========
data "consul_service" "vault" {
  name = "vault"
}

# ==========
# providers
# ==========
provider "consul" {
  version = "~> 2.2"
}
provider "vault" {
  version = "~> 1.4"
  address = "https://${data.consul_service.vault.service.0.address}:${data.consul_service.vault.service.0.port}"
}
```

As you can see, I'm leveraging the data source to get the address of the service.

Because it's the catalog interface, it returns all nodes, regardless of health.

This works brilliantly when there's no dead nodes, but breaks the moment there's a dead node, which is quite common if you have client nodes in an autoscale group or similar.

[...]

PR #89, opened 2019-01-29
Add the consul_service_health data source

remilapeyre:
This PR adds a new data source `consul_service_health` that fetch checks information about a given service.

See #87


remilapeyre:
I wish I would be able to export attributes as sub-resources so they would be accessible like so:

```
${data.consul_service_health.vault.nodes.0.service.id}
${data.consul_service_health.vault.nodes.0.service.name}
${data.consul_service_health.vault.nodes.0.service.tags}
${data.consul_service_health.vault.nodes.0.service.address}
```

but it seems to me that this is not possible so I had to resort to an ugly `_`:

```
${data.consul_service_health.vault.nodes.0.service_id}
${data.consul_service_health.vault.nodes.0.service_name}
${data.consul_service_health.vault.nodes.0.service_tags}
${data.consul_service_health.vault.nodes.0.service_address}
```

remilapeyre:
To correctly test this PR, it is better to wait for the merge of #64

paultyng:
Will this let you wait until a service is healthy? For example if you just created a new service in the TF script and wanted to wait for it to be fully available?

consul/resource_consul_service.go
```diff
@@ -497,11 +497,18 @@ func retrieveService(client *consulapi.Client, name string, ident string, node s
 	// Only one service with a given ID may be present per node
 	for _, s := range services {
 		if (s.ServiceID == ident) && (s.Node == node) {
+			// Fetch health-checks for this service
 			healthChecks, _, err := client.Health().Checks(name, &qOpts)
 			if err != nil {
 				return nil, fmt.Errorf("Failed to fetch health-checks: %v", err)
 			}
-			s.Checks = healthChecks
```
pearkes: Could this cause backwards incompatibility issues? By returning less checks than expected?

website/docs/d/service_health.html.markdown
```diff
@@ -0,0 +1,100 @@
...
+description: |-
+  Filter service instances based on health status
+---
+
+# consul_service_health
+
+`consul_service_health` can be used to get the list of the instances that
+are currently healthy, according to their associated  health-checks.
+The result includes the list of service instances, the node associated to each
+instance and its health-checks.
```
pearkes: I think we should add a few lines here about this being a resource that will likely change frequently as service health changes (theoretically) quite a lot, and that you can expect different results on a frequent basis.

consul/resource_consul_service.go
remilapeyre: It's true that there will be less health-checks but the previous behaviour was returning health-checks not associated with this service instance, I think it was a bug, people relying on this could not make the correct decisions could they?

Issue #95, opened 2019-04-05
Find a way to bootstrap Consul ACL support


remilapeyre:
https://github.com/terraform-providers/terraform-provider-consul/pull/60 introduces support for ACL with `consul_acl_policy` and `consul_acl_token`. Both require the ACL to have been previously bootstrapped manually to work.

Bootstrapping ACLs requires to keep track of the master token given by https://www.consul.io/api/acl/acl.html#bootstrap-acls so it would be written as clear text in the terraform state.
The change in ACLs in Consul 1.4 was made so that the secret IDs would not need to be saved in unsecure situation and accessor IDs could be used instead.

Once ACLs are bootstrapped, the operator also needs to update the provider configuration and to set `token` or set the `CONSUL_HTTP_TOKEN` environment variable.

It would be nice to have a secure way to automate this.

Related: https://github.com/hashicorp/terraform/issues/9556

frederikbosch:
Is there someone aware of a valid workaround? `data "external"{}` is re-run every time. Bootstrapping only needs to be executed once.

remilapeyre:
Hi @frederikbosch, we don't have a satisfactory way to securely bootstrap the ACL support without the master token ending up in the Terraform state in plaintext.

Storing secrets in the Terraform state is a complicated issue, you can read more about it here : https://github.com/hashicorp/terraform/issues/516

 > Bootstrapping only needs to be executed once

Did you try using a provisioner (https://www.terraform.io/docs/provisioners/local-exec.html) ?

frederikbosch:
@remilapeyre Exactly, I understand. I used `local-exec` indeed, combined with `remote-exec` to create policies, tokens and apply directly on the remote machines. I did that for Consul, Vault and Nomad. This did have the downside that the tokens ended-up in my (encrypted) remote S3 state. This is not a preferred solution, as indicated in the linked thread. However, manual bootstrapping is also not a preferred activity as alternative. I hope a better solution will become available soon.

remilapeyre:
Hi everyone, I finally found a way to bootstrap the ACL system without having to write the management token in the Terraform state. https://github.com/hashicorp/vault/pull/10751 makes it possible to use Vault to do the ACL bootstrapping securely. Once it has been merged in Vault I will write a guide to show how to do this and generate additional Consul token from Vault using Terraform.

lethargosapatheia:
[...]

PR #98, opened 2019-04-12
Add computed secret_id to the consul_acl_token attributes

xeivieni:
In order to store the token generated into vault for example.


remilapeyre:
Hi @xeivieni, thanks for taking the time to write a pull-request! Not having the secret token accessible through the resource was deliberate as doing so would save it in plaintext in the terraform state, defeating the purpose of saving it in Vault.

Not having to save the secret token in non-secure places was the point of splitting them as Access token / Secret token in the first place. 

In some cases, the need of recording secret tokens can be avoided altogether. It would be helpful for me that you add your use-case to https://github.com/terraform-providers/terraform-provider-consul/issues/95 so that I can see if another solution exists.

PR #102, opened 2019-04-22
Fix update of consul_acl_token

remilapeyre:
I made a mistake during the review of #60, updating tokens currently crashes Terraform because of the wrong type assertion at https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_acl_token.go#L124.

### DIFF
--- a/consul/resource_consul_acl_token.go
+++ b/consul/resource_consul_acl_token.go
@@ -14,6 +14,9 @@ func resourceConsulACLToken() *schema.Resource {
 		Read:   resourceConsulACLTokenRead,
 		Update: resourceConsulACLTokenUpdate,
 		Delete: resourceConsulACLTokenDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"description": {
--- a/consul/resource_consul_acl_token.go
+++ b/consul/resource_consul_acl_token.go
@@ -24,7 +24,6 @@ func resourceConsulACLToken() *schema.Resource {
 			"policies": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				ForceNew: true,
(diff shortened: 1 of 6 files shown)


consul/resource_consul_acl_token.go
```diff
@@ -24,7 +24,6 @@ func resourceConsulACLToken() *schema.Resource {
 			"policies": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				ForceNew: true,
```
pearkes: Is this intentionally included here? I'm not seeing the intent to change this behavior. Maybe it was test related?

consul/resource_consul_acl_token_test.go
```diff
@@ -38,12 +38,18 @@ func TestAccConsulACLToken_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("consul_acl_token.test", "description", "test"),
 					resource.TestCheckResourceAttr("consul_acl_token.test", "policies.#", "1"),
+					resource.TestCheckResourceAttr("consul_acl_token.test", "policies.1785148924", "test"),
 					resource.TestCheckResourceAttr("consul_acl_token.test", "local", "true"),
 				),
 			},
 			{
-				Config:  testResourceACLTokenConfigBasic,
```
pearkes: Should we be retaining this test case? Maybe after the update?

consul/resource_consul_acl_token.go
remilapeyre: This was making it impossible to update the policies associated to a token without changing its Access ID and Secret ID. I don't think this is what was expected, especially since there was no test for this.

Pinging @bhuisgen, as the original author he may have more context regarding this.

pearkes: I'm okay with merging as-is but it'd be good to track down the reason there. I think this makes sense to me though.

Issue #101, opened 2019-04-19
consul_service resource overrides node meta


ashald:
It's clear from the source code that we can see at https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_service.go#L321-L327:
```go
	var nodeMeta map[string]string
	nodeMeta = make(map[string]string)
	if d.Get("external").(bool) {
		nodeMeta["external-node"] = "true"
		nodeMeta["external-probe"] = "true"
	}
	registration.NodeMeta = nodeMeta
```

### Affected Resource(s)
Please list the resources as a list, for example:
- `consul_service`

### Expected Behavior
At the bare minimum, one would expect that node meta would be preserved.

On top of that, I believe `external-node` and `external-probe` should not be overridden if already defined. For instance, we need to disable the ESM probe while keeping ESM health checks active so we set `external-probe = false` on the node.

remilapeyre:
Thanks for the report @ashald, I overlooked this when adding `external-node` and `external-probe`. What you propose makes sense.

ashald:
I'm pretty sure I'd do the same if I were in your shoes implementing those resources. There is a lot of complexity to it and it's easy to be tricked by that.

Thanks for your prompt attention!

remilapeyre:
Thanks for your feedback, after more look into it, it seems that `consul_node` was not detecting changes made to the `address` and `meta` attributes which led to the bug in the first place. This is fixed in #104 that should be merged in the next few days.

In the meantime, you can work around the bug by removing `external` from your Terraform scenario and set `external-probe` and `external-node` manually in Consul node. It will be working as well as it was before (set the meta information but won't detect changes to it).

Once #104 is merged, it will start detecting external changes without any new action from your part.

ashald:
Thanks for the update!

PR #104, opened 2019-04-26
Fix consul_node not detecting changes

remilapeyre:
`consul_node` were not updating their `address` and `meta` attributes during plans. This let another bug go unnoticed where `consul_service` could change `consul_node` attributes when `external` would be set to `true` (see https://github.com/terraform-providers/terraform-provider-consul/issues/101).

Multiple services could have conflicting configuration:

``` terraform
resource "consul_node" "compute" {
	name    = "compute-example"
	address = "www.hashicorptest.com"
}

resource "consul_service" "example1" {
	name = "example"
	node = "${consul_node.compute.name}"
	port = 80

	external = true
}


resource "consul_service" "example2" {
	name = "example"
	node = "${consul_node.compute.name}"
	port = 80

	external = true
}

// What should be the node's meta.external-probe and meta.external-node?
```

This patch fix `resourceConsulNodeRead()` and deprecate `external` from `consul_service`.
Another option would have been move `external` to the `consul_node` resource but it would then have conflicted with the `meta` attribute. We could use a custom `DiffSuppressFunc` to avoid this but I'm not sure it is worth the maintenance cost.

### DIFF
--- a/consul/resource_consul_node.go
+++ b/consul/resource_consul_node.go
@@ -129,8 +129,15 @@ func resourceConsulNodeRead(d *schema.ResourceData, meta interface{}) error {
 
 	if n == nil {
 		d.SetId("")
+		return nil
 	}
 
+	if err = d.Set("address", n.Node.Address); err != nil {
+		return fmt.Errorf("Failed to set 'address': %v", err)
+	}
+	if err = d.Set("meta", n.Node.Meta); err != nil {
+		return fmt.Errorf("Failed to set 'meta': %v", err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


Shaiou:
Hi, any news as to when this will be merged ?

remilapeyre:
Hi @Shaiou, thanks for the ping ! We should be able to cut a new release next week :)

Shaiou:
Hi, any news regarding the merge ?

Shaiou:
:+1:

Shaiou:
Sorry to bother again, is there going to be a release/tag soon ?

Issue #57, opened 2018-07-30
consul_node resource doesn't inherit datacenter attribute from provider definition when defined


splashx:
### Terraform Version
Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Affected Resource(s)

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Terraform Configuration Files

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Expected Behavior

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Actual Behavior

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Steps to Reproduce

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### Important Factoids

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

### References

Refer to https://github.com/terraform-providers/terraform-provider-consul/issues/56

remilapeyre:
Hi @splashx , thanks for reporting the bug.

In #56 you say that `Permission Denied` is expected and you are trying to register an external service.
Do I understand correctly that:
  - you want to default to the configuration of the provider when `/agent/self` returns `Permission Denied`
  - the fact that this node is external is not really important and the bug would manifest itself as long as the token used to call Consul API has not the permission to call `/agent/self`

?

remilapeyre:
I thought we had to do

1. Datacenter of the resource if configured
2. Datacenter of the agent if the one in the resource is not defined
3. Datacenter of provider as last resort

Currently, it seems to me, that the behavior is 1 and 2 and never 3 but reading the documentation (https://github.com/terraform-providers/terraform-provider-consul/blame/master/website/docs/index.html.markdown#L57 and https://github.com/terraform-providers/terraform-provider-consul/blame/master/website/docs/r/keys.html.markdown#L40-L41) it should actually be:

1. Datacenter of the resource if configured
2. Datacenter of provider if set
3. Datacenter of the agent as of last resort

Unless I'm missing something, this is not what is being done in https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_keys.go#L330-L340

From a quick glance at the code, most of the resources are using `getDC` to get the `datacenter` when it is not defined in the resource except `consul_intentions` and `consul_service`.

[...]

PR #105, opened 2019-04-26
Use provider datacenter as default when the agent cannot be reached

remilapeyre:
According to the current documentation, the datacenter attribute should
default to the provider configuration and if it is not set to the
datacenter of the agent terraform is connected to:

	* `datacenter` - (Optional) The datacenter to use. This overrides the datacenter in the
	  provider setup and the agent's default datacenter.

This is not what the code actually does and the datacenter in the provider
configuration is not used most of the time.

Since changing the plugin behavior to what is actually documented may
break code relying on the current one. To keep backward compatibility,
we can change the documentation to:

	* `datacenter` - (Optional) The datacenter to use. This overrides the
	  agent's default datacenter and the datacenter in the provider setup.

and fall back to the provider configuration only when the datacenter of
the agent to which terraform is connected to can not be read (when ACL
is enabled and the token used by terraform does not have permission to
read /agent/self for example).

The change is very noisy as it requires to change the `ConfigureFunc` of
the provider from `(*consulapi.Client, error)` to `(*Config, error)` and
update all places where was being used.

The interesting change is at https://github.com/terraform-providers/terraform-provider-consul/pull/105/files#diff-7c162fe0d0220927a53a0f1977a47a71R361

Discussion: https://github.com/terraform-providers/terraform-provider-consul/issues/57

### DIFF
--- a/consul/data_source_consul_agent_config.go
+++ b/consul/data_source_consul_agent_config.go
@@ -2,7 +2,7 @@ package consul
 
 import (
 	"fmt"
-	consulapi "github.com/hashicorp/consul/api"
+
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -51,7 +51,10 @@ func dataSourceConsulAgentConfig() *schema.Resource {
 }
 
 func dataSourceConsulAgentConfigRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*consulapi.Client)
+	client, err := meta.(*Config).Client()
+	if err != nil {
+		return err
+	}
 	agentSelf, err := client.Agent().Self()
 	if err != nil {
 		return err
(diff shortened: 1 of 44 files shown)


consul/data_source_consul_agent_config.go
```diff
@@ -51,7 +51,10 @@ func dataSourceConsulAgentConfig() *schema.Resource {
 }
 
 func dataSourceConsulAgentConfigRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*consulapi.Client)
+	client, err := meta.(*Config).Client()
+	if err != nil {
+		return err
```
pearkes: Should we prefix these errors? I wonder if it will be obvious what caused them.

remilapeyre: I put those error checks to cover all possibilities but this should never happen since the config is already checked during the provider configuration at https://github.com/terraform-providers/terraform-provider-consul/pull/105/files#diff-b338f46ba92504078fcef3e0e3a1b2cbR127. I could add a helper function to make this more obvious:

```
func getClient(meta interface{}) consulapi.Client {
    // ignore error as the configuration has already been checked in providerConfigure()
    client, _ := meta.(*Config).Client()
    return client
}
```
and replace all the 
```
	client, err := meta.(*Config).Client()
	if err != nil {
		return err
	}
```
by 
```
     client := getClient(meta)
```

PR #110, opened 2019-05-17
Export token secret and add ability to specify dc/token for ACL provisioning

andrewstucki:
So, I followed form from what I found in `query_options.go` and expanded it to encompass the ability to set `WriteOptions` for the consul api requests. The `getRequestOpts` function differs a good amount from `getQueryOpts` because, as far as I can tell, `getQueryOpts` appears to be broken (it looks like it's always going to have default values in the `QueryOptions` structure due to deserializing `ResourceData` improperly).

~~Also, wasn't sure exactly how acceptance testing was generally done with this repo/who pays attention to it since it appears 1) to be off in CI (no `make testacc` invocation) and 2) to be broken on the docker-compose setup due to the ACL system not being enabled.~~

~~I tested this by adding a configuration file for docker to pick up with ACLs enabled and then ran the acceptance tests. If you want me to try and fix up CI/the testing setup I can, but, as noted, not sure if anyone seems to care about acceptance testing.~~

edit:

re-read the README and saw that the recommended way for acceptance testing is booting up a local consul rather than using the docker-compose stuf

### DIFF
--- a/consul/request_options.go
+++ b/consul/request_options.go
@@ -2,7 +2,6 @@ package consul
 
 import (
 	"fmt"
-	"log"
 	"strings"
 
 	consulapi "github.com/hashicorp/consul/api"
@@ -65,8 +64,6 @@ func getRequestOpts(d *schema.ResourceData, client *consulapi.Client) (*consulap
 		requestOpts.Datacenter = dc
 		queryOpts.Datacenter = dc
 	}
-
-	log.Printf("[DEBUG] Options %+v %+v", d, requestOpts)
 
 	return requestOpts, queryOpts, nil
 }
--- a/consul/request_options.go
+++ b/consul/request_options.go
@@ -1,0 +1,85 @@
+package consul
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+const (
+	requestOptions = "request_options"
+
+	requestOptDatacenter = "datacenter"
+	requestOptToken      = "token"
+)
+
+var schemaRequestOpts = &schema.Schema{
+	Optional: true,
+	Type:     schema.TypeSet,
+	Elem: &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			requestOptDatacenter: {
+				// Optional because we'll pull the default from the local agent if it's
+				// not specified, but we can query remote data centers as a result.
+				Optional: true,
+				Type:     schema.TypeString,
+			},
+			requestOptToken: {
+				Optional:  true,
+				Type:      schema.TypeString,
+				Sensitive: true,
+			},
+		},
+	},
+}
+
+func getRequestOpts(d *schema.ResourceData, client *consulapi.Client) (*consulapi.WriteOptions, *consulapi.QueryOptions, error) {
+	requestOpts := &consulapi.WriteOptions{}
+	queryOpts := &consulapi.QueryOptions{}
+
+	if v, ok := d.GetOk(requestOptions); ok {
+		options := v.(*schema.Set).List()
+		if len(options) > 0 {
+			opts := options[0].(map[string]interface{})
+
+			if token, ok := opts[requestOptToken]; ok {
+				requestOpts.Token = token.(string)
+				queryOpts.Token = token.(string)
+			}
+
+			if dc, ok := opts[requestOptDatacenter]; ok {
+				requestOpts.Datacenter = dc.(string)
+				queryOpts.Datacenter = dc.(string)
+			}
+		}
+	}
+
+	if requestOpts.Datacenter == "" {
+		dc, err := getLocalDC(client)
+		if err != nil {
+			return nil, nil, err
+		}
+		requestOpts.Datacenter = dc
+		queryOpts.Datacenter = dc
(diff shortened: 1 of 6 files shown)


remilapeyre:
Hi @andrewstucki, there has been a decision with the Consul team to not expose the Secret IDs in Terraform, tokens were split into AccessID / SecretID so there would no need to write the Secret IDs in unsecure locations. This PR would write all Secret IDs in the Terraform state in plaintext which is something we want to avoid.

Could you give more details about your use-case? Why do you need the Secret IDs (for programmatic access, user access, other... ) ?

andrewstucki:
@remilapeyre my general use-case is for generating tokens and policies on a per-service basis and injecting them into configuration via terraform.

Having used terraform for a number of years, there are already a myriad of places in most providers where, if you leverage terraform heavily enough, you're going to need to have solid controls around access to terraform state because sensitive values are written to disk all the time. For example, if you use AWS and RDS, the [rds password](https://github.com/terraform-providers/terraform-provider-aws/blob/c75eae11582a51fe94f3544fd23696c646379c0f/aws/resource_aws_db_instance.go#L54) is written into the state file.

Without this change in the provider it actually becomes pretty cumbersome to add ACL tokens via this provider. You have to generate the token, then manually retrieve it to reference it and hardcode the value somewhere to actually use it. The only alternative I can think of aside from exposing this as part of the resource is potentially adding an ACL token data source so that no secrets are written to disk unless you specifically use the new data source.

The second piece of functionality this addresses is also specifying which token is used in the request when attempting to create policies and tokens. Here's a general sketch of the use-case:

## Desired

```
resource "random_uuid" "master_token" {}

resource "aws_instance" "consul" {
  ...

  user_data = <<-EOF
  #!/bin/bash
  bootstrap-consul "${random_uuid.master_token.result}"
  EOF
}

output "master_token" {
  value = "${random_uuid.master_token.result}"
  sensitive = true
}
```

and in another manifest that pulls this in using `terraform_remote_state`

```
data "terraform_remote_state" "consul" {
  ...
}

resource "consul_acl_policy" "app" {
  ...
  request_options {
    token = "${data.terraform_remote_state.consul.master_token}"
  }
}

resource "consul_acl_token" "app" {
  ...
  policies    = ["${consul_acl_policy.app.name}"]
  request_options {
    token = "${data.terraform_remote_state.consul.master_token}"
  }
}

resource "aws_instance" "app" {
  ...

  user_data = <<-EOF
  #!/bin/bash
  bootstrap-app "${consul_acl_token.app.secret}"
  EOF
}
```

## Currently

The first manifest would be the same:

```
resource "random_uuid" "master_token" {}

resource "aws_instance" "consul" {
  ...

  user_data = <<-EOF
  #!/bin/bash
  bootstrap-consul "${random_uuid.master_token.result}"
  EOF
}

output "master_token" {
  value = "${random_uuid.master_token.result}"
  sensitive = true
}
```

Now you'd have to break the second step of provisioning into multiple steps. First, output the `master_token` and export it using `CONSUL_TOKEN` environment variables or check it in to a variables file and pass it to the provider configuration block (this potentially also makes it almost impossible to work with different tokens for different resources). Next, run:

```
resource "consul_acl_policy" "app" {
  ...
}

resource "consul_acl_token" "app" {
  ...
  policies    = ["${consul_acl_policy.app.name}"]
}
```
Retrieve the created token from this last step **_manually_** from the UI. In a _**third**_ manifest do something like the following after writing the token into a `.tfvars` file:

```
variable "app_token" {}

resource "aws_instance" "app" {
  ...

  user_data = <<-EOF
  #!/bin/bash
  bootstrap-app "${var.app_token}"
  EOF
}
```

Provisioning an entire environment manually copying and pasting generated tokens around to environment variables/variable files then becomes 1) time intensive, and 2) error prone and fairly at odds with the whole idea of infrastructure as code.

If you have any other ideas, let me know.

andrewstucki:
Any updates/thoughts?

remilapeyre:
Thanks for your patience @andrewstucki, after reflexion I think we will add a new datasource to fetch the secret ID so that users that do not use the datasource won't have the secret ID stored in the TF state while those who need it still can retrieve it.

andrewstucki:
@remilapeyre ok, I'm fine with that, I'll work on a PR in the next day or two to add the data source. Would you be amenable to me splitting out the `request_options` piece of this PR into something reviewable? Any thoughts about the naming conventions?

remilapeyre:
What do you think of:

```
data "consul_acl_secret_id" "test" {
    access_id = "..."
    datacenter = "dc1"
    pgp_key = "..."
    token = "..."
}
```

with only `access_id` being required and it exports `secret_id`?

For the coding style, you don't need to extract all labels in constants, you can keep them inlined like in https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_service.go

remilapeyre:
Thanks for opening this PR @andrewstucki !

Regarding the bug in `query_options` it has been fixed in fb70af9b16eb277c9a4b167bf95ff750458f4d45.

For accessing the secret id of a token, we prefer to use a separate datasource so those that actually need it can use this data source and those who don't won't have the secret ID saved in their Terraform state.

For setting `datacenter` and `token` in `consul_acl_policy` and `consul_acl_token`, having both attributes at the root of the resource will be more consistent the other resources.

PR #114, opened 2019-05-21
added warning so users are aware of the prerequisite

ars05:
Adding the supported version to the description prevents users to raise issue if they are using an older version of Consul and Terraform which doesn't provide the feature.

### DIFF
--- a/website/docs/r/acl_policy.html.markdown
+++ b/website/docs/r/acl_policy.html.markdown
@@ -8,9 +8,8 @@ description: |-
 
 # consul_acl_policy
 
-The `consul_acl_policy` resource writes an ACL policy into Consul.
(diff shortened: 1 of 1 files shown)


remilapeyre:
This is quite a large warning, what do you think about just changing the phrase 

> The `consul_acl_policy` resource writes an ACL policy into Consul.

to something like

> Starting with Consul 1.4, the `consul_acl_policy` can be used to managed [Consul ACL policies](https://www.consul.io/docs/acl/acl-system.html#acl-policies).

ars05:
Sound good to me. Removed the warning, added the version to the description.

remilapeyre:
Thanks @ars05 !

PR #117, opened 2019-05-28
Mark local attribute of consul_acl_token as ForceNew

greut:
It's more a question than a real PR.

Shouldn't the `local` attribute be set to `ForceNew`? It fails witha 500 or crashes doing `panic: interface conversion : interface {} is *schema.Set, not []interface {}` in `consul.resourceConsulACLTokenUpdate` (line 124) using the provider 2.3.0_x4 and a Consul 1.5.1 server.

```text
=== RUN   TestAccConsulACLToken_basic
--- FAIL: TestAccConsulACLToken_basic (0.06s)
    testing.go:538: Step 1 error: Error applying: 1 error occurred:
                * consul_acl_token.test: 1 error occurred:
                * consul_acl_token.test: error updating ACL token "d56b4d5f-ed65-3ed4-2064-94935b5bf129": Unexpected response code: 500 (cannot toggle local mode of d56b4d5f-ed65-3ed4-2064-94935b5bf129)
```

Cheers,


remilapeyre:
Hi @greut, thanks for noticing that, the crash was fixed on #102 but according to the documentation for the `local` parameter (https://www.consul.io/api/acl/tokens.html#local-1):

> This value must match the existing value or the request will return an error.

so your PR seems good.

greut:
@remilapeyre that was quick, thanks a lot!

PR #118, opened 2019-05-29
Add support for Terraform 0.12

remilapeyre:


consul/resource_consul_prepared_query_test.go
```diff
@@ -163,6 +163,10 @@ func testAccConsulPreparedQueryNearestN(t *testing.T) func() {
 	}
 }
 
+func testAccCheckConsulPreparedQueryAttrValue(attr, val string) resource.TestCheckFunc {
```
bflad: This new function appears to be unused.

consul/resource_consul_keys.go
```diff
@@ -19,6 +19,13 @@ func resourceConsulKeys() *schema.Resource {
 		SchemaVersion: 1,
 		MigrateState:  resourceConsulKeysMigrateState,
 
+		CustomizeDiff: func(d *schema.ResourceDiff, _ interface{}) error {
```
bflad: Was this change intended for this pull request?

remilapeyre: Yes, this is part of ff2785a7 and related to https://github.com/hashicorp/terraform/issues/21070.

I think (but I'm not sure) that it is because Terraform 0.12 now expect the result of the apply to match the plan so updating the SDK highlighted a bug that was previously silent.

remilapeyre:
Hi @bflad, thanks for the review. It should be good to go now.

Issue #112, opened 2019-05-21
Meta block on consul_service


StuporHero:
In connection to #101, I'm wondering if there is a technical reason we can't implement a meta block for `consul_service`. I'm trying to register details of my Oracle database system created in OCI that could lose support if I install a consul agent directly on it, so I'm treating it like an external service. I'd like to use the specifics of the database in a provisioning workflow that clones a database and updates the config file on the middleware to contain an entry for the new database. I'd be happy to try my hand at implementing it, but I want to make sure this is something that is both in line with the goals of the project and isn't waiting on something else before I make the time investment.

remilapeyre:
Hi @StuporHero, thanks for opening this issue!

`Meta` is optional and I think nobody needed it until now, adding a `meta` attribute to `consul_service` would be nice. You can try yo implement it and I would be happy to guide you through the process if you need help.

StuporHero:
@remilapeyre Actually, I just reviewed #104, and it looks like that PR adds the `meta` block. Merging that PR should get me what I need. Any idea when that'll be possible? It looks like it was waiting on a release that may have been delayed.

remilapeyre:
Hi @StuporHero, I think you are confusing the `NodeMeta` and `ServiceMeta` attributes. Both can be set when registring a service in Consul (see https://www.consul.io/api/catalog.html#sample-response-3). `NodeMeta` is exposed in `consul_node` but `ServiceMeta` is not present in `consul_service` and there is no way to set it as far as I know.

I think #104 should be merged in a couple of days.

StuporHero:
@remilapeyre You would be correct. I don't know what I saw that made me think #104 would fix my problem.

StuporHero:
@remilapeyre I've got some questions about writing the tests. Is it okay if I go ahead and create a pull request with what I've got so far to make it easier to collaborate on it?

remilapeyre:
Yes absolutely!

remilapeyre:
Thanks @StuporHero for the PR!

PR #122, opened 2019-05-31
Adding meta to consul_service

StuporHero:
Solves #112 @remilapeyre I can't seem to get the tests to fail. I can't seem to figure out what they're actually doing. I'll add comments to lines to explain what I thought was true but obviously isn't.

### DIFF
--- a/consul/resource_consul_service_test.go
+++ b/consul/resource_consul_service_test.go
@@ -26,13 +26,12 @@ func TestAccConsulService_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("consul_service.example", "port", "80"),
 					resource.TestCheckResourceAttr("consul_service.example", "tags.#", "1"),
 					resource.TestCheckResourceAttr("consul_service.example", "tags.0", "tag0"),
-					resource.TestCheckResourceAttr("consul_service.example", "meta.test", "test"),
+					resource.TestCheckResourceAttr("consul_service.example", "meta.%", "0"),
 					testAccConsulExternalSource,
 				),
 			},
 			{
-				PreConfig: testAccRemoveConsulService(t),
-				Config:    testAccConsulServiceConfigBasic,
+				Config: testAccConsulServiceConfigBasicMeta,
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("consul_service.example", "id", "example"),
 					resource.TestCheckResourceAttr("consul_service.example", "service_id", "example"),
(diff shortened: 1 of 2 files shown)


consul/resource_consul_service_test.go
```diff
@@ -26,6 +26,7 @@ func TestAccConsulService_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("consul_service.example", "port", "80"),
 					resource.TestCheckResourceAttr("consul_service.example", "tags.#", "1"),
 					resource.TestCheckResourceAttr("consul_service.example", "tags.0", "tag0"),
+					resource.TestCheckResourceAttr("consul_service.example", "meta.test", "test"),
```
StuporHero: I added this believing it would be checked, but when I set the value to `fail`, the test still passes. What exactly is this whole portion doing?

consul/resource_consul_service_test.go
```diff
@@ -311,12 +314,15 @@ resource "consul_service" "example" {
 	node    = "${consul_node.compute.name}"
 	port    = 80
 	tags    = ["tag0"]
-  }
+	meta    = {
```
StuporHero: This goes along with the previous comment. I'm pretty sure understanding this will be enough for me to make the rest of this test actually test my change.

remilapeyre:
I took a quick look at the code, it seems good, you can omit `ForceNew` from the schema as it defaults to `false`.

Are you runnning the tests with `make test`?

Tests of Terraform providers are usually separated in two sets, some unit tests and some acceptance tests. Since acceptance tests create some resources, they are not run by default.

To run them, you will need a Consul server with the correct configuration, you can use `consul agent -dev -config-file ./consul_test.hcl`, and then you can run `make testacc` with `CONSUL_HTTP_ADDR` and the `CONSUL_HTTP_TOKEN` variables set. `CONSUL_HTTP_ADDR=http://localhost:8500 CONSUL_HTTP_TOKEN=master-token make testacc` should do the trick.

StuporHero:
@remilapeyre I was using `go test`, but `make test` doesn't seem to do anything. I just get a log like the following:
```bash
make[1]: Entering directory '~/go/src'
make[1]: *** No rule to make target 'testacc'.  Stop.
make[1]: Leaving directory '~/go/src'
```
I'm also getting the following when I try `make fmt` as suggested by the travis-ci log:
```bash
make: *** No rule to make target 'fmt'.  Stop.
```
I'm using GNU Make 4.2.1 for Windows if that makes a difference.

remilapeyre:
Are you in the correct folder? Should should see a `Makefile` when running `ls`.

StuporHero:
@remilapeyre Yes, I was in the wrong folder. I also had the wrong tree structure. I've corrected it and am now using `make test`, but I'm still not able to make the test fail.

StuporHero:
@remilapeyre Ah, jeez. I really should have read the README.md first. I got it working right. Thanks for your patience with me. I'm normally better than this. With that, this PR should be ready to merge unless anyone else has objections.

consul/resource_consul_service_test.go
StuporHero: If I had read the README.md, I would have understood that the test does do what I thought it did but only if there is a consul server available for the tests to run against.

remilapeyre:
Thanks for the help @StuporHero ! Other than some cosmetic modifications, I just added the part to read the service meta information in `resourceConsulServiceRead()`: https://github.com/terraform-providers/terraform-provider-consul/pull/122/files#diff-a4e9ad12519f0ecc54030a9237aa517fR397.

PR #127, opened 2019-06-27
Add the config_entry resource

remilapeyre:


website/docs/r/configuration_entry.markdown
```diff
@@ -57,13 +57,15 @@ proxies. Can only be set when `kind` is `proxy-defaults`.
...
+* `id` - The id of the configuration entry.
+
+* `kind` - The kind of the configuration entry, `proxy-defaults` or
 `service-defaults`.
 
-* `name` - (Required) The name of the configuration entry.
+* `name` - The name of the configuration entry.
 
-* `config` - (Optional) A map of configuration values.
+* `config` - (A map of configuration values.
```
greut:
```diff
- (A map
+ A map
```

remilapeyre: Thanks @greut !

consul/resource_consul_configuration_entry.go
```diff
@@ -0,0 +1,176 @@
...
+			"config": {
+				Type:     schema.TypeMap,
+				Optional: true,
+			},
+
+			"token": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+			},
```
banks: Hmm I see why mixing all the fields in every config Entry type we add gets a bit messy here. This is fine for now but I expect the number of fields here to increase a lot as we add more options to Connect etc.

Is there a way we can make this more generic so it doesn't need constant maintenance to keep up with the config syntax for all these types?

For example in Consul core config we allow bootstrapping these from HCL but we internal just type it as `map[string]interface{}` so we don't have to maintain the config entry schema in lots of different places.

I'm not super familiar with TF providers though so I'm not sure what is possible and what the downsides are of making it more generic. What do you think @remilapeyre ?

banks: For example, the beta release we will cut this week adds 3 more "kinds" here that have another 10+ top level fields we'd have to add to this schema my hope was we could keep this generic and then have it just work for new types automatically.

banks: Is it possible for the top-level Schema just to be a `schema.TypeMap`? Or to define kind and name but then have TF just present all the other fields generically?

consul/resource_consul_configuration_entry.go
```diff
@@ -0,0 +1,176 @@
...
+			// The config entry has been removed
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Failed to fetch '%s' config entry: %#v", configName, err)
+	}
+
+	switch configKind {
+	case consulConfigEntryProxyDefaults:
+		if err = d.Set("config", configEntry.(*consulapi.ProxyConfigEntry).Config); err != nil {
```
banks: I'd be tempted to type assert this with, `entry, ok := configEntry.(*consulapi.ProxyConfigEntry)` and return an error on bad type rather than panic and crash. It is a bug for sure but panic and crash is not a great UX if we ever do return something bad or the response is garbled by a proxy or similar.

consul/resource_consul_configuration_entry.go
```diff
@@ -0,0 +1,176 @@
...
+		if configName != "global" {
+			return fmt.Errorf("'name' must be 'global' when 'kind' is 'proxy-defaults'")
+		}
+		if d.Get("protocol").(string) != "" {
+			return fmt.Errorf("'protocol' must not be set when 'kind' is 'proxy-defaults'")
+		}
+	case consulConfigEntryServiceDefaults:
+		if len(d.Get("config").(map[string]interface{})) != 0 {
+			return fmt.Errorf("'config' must not be set when 'kind' is 'service-defaults'")
+		}
```
banks: This validation is great but I guess if we want this to be generic it would have to go. I wonder if it makes the UX bad if we just accept anything and rely on Consul reporting errors when we attempt to write it on apply.

I can see how having earlier feedback is nicer I just worry that the validation here will end up being a burden to keep in sync with the equivalent in Consul - for example when we add namespaces there will be valid `proxy-defaults` configs other than `global` as you'll be able to set one per namespace but anyone with the old provider won't work.

website/docs/r/configuration_entry.markdown
```diff
@@ -0,0 +1,71 @@
+---
+layout: "consul"
+page_title: "Consul: configuration_entry"
+sidebar_current: "docs-consul-resource-configuration-entry"
+description: |-
+  Registers a configuration entry in Consul.
+---
+
+# consul_configuration_entry
```
banks: can we name it `consul_config_entry`? We tend to try to use the full word "configuration" in documentation etc as a style thing, but we abbreviated this name in all of the places that use it as an identifier for this resource type. For example our API is `/v1/config/...` and the CLI command is `consul config ...`

website/docs/r/configuration_entry.markdown
```diff
@@ -0,0 +1,71 @@
...
+`proxy-defaults` or `service-defaults`.
+
+* `name` - (Required) The name of the configuration entry being registred. If
+`kind` is `proxy-defaults`, `name` must be `global`.
+
+* `config` - (Optional) An arbitrary map of configuration values used by Connect
+proxies. Can only be set when `kind` is `proxy-defaults`.
+
+* `protocol` - (Optional) The protocol of the service. Can only be set when
+`kind` is `service-defaults`.
```
banks: I suggest if we can make this generic that we just document the kind and name and then link to the https://www.consul.io/docs/agent/config_entries.html#configuration-entry-kinds docs for details so we don't have to remember to keep this in sync too across new kinds.

remilapeyre:
Hi @banks, I've been trying to use a more generic `config` attribute in the Terraform schema but get stuck because Terraform wants a strict schema so I could find nothing that works for every ConfigEntry types.

What if we just use a `config` string field in which the user can write arbitrary HCL that we parse and give to `structs.DecodeConfigEntry`?

For example : 

``` hcl
resource "consul_config_entry" "service-router" {
  kind = "service-router"
  name = "web"
  
  config = <<-EOF
    routes = [
      {
        match {
          http {
            path_prefix = "/admin"
          }
        }

        destination {
          service = "admin"
        }
      }
    ]
  EOF
}
```

This could make it work albeit the integration with Terraform and user-experience might not be the best.

banks:
Sorry it took so long to respond @remilapeyre - HashiConf was upon us last week!

I'm honestly not sure about that UX - I think it would be good to loop in some Terraform folks here to steer us towards the best path. There may well be good reasons not to do the embedded HCL route beyond just looking a bit ugly that are too subtle for my experience to tease out.

I'll ask around internally first and see what the consensus is.

banks:
Hey @remilapeyre I have spoken to our Terraform Gurus (@paultyng) and have some good ideas now.

## The Future

First up the TF providers team are trying to solve this very problem for Kube right now - 0.12 dynamic types make it possible in the language to do much better but the SDK to support that in providers is still work in progress.

Even when it does land, it will only work for 0.12 so keep an eye out for a major SDK update soon but we need another solution medium term for 0.11 support anyway.

## The Present

The present prior art here is pretty much what you came up with. You probably know this but for the purpose of future readers, Nomad job files can inline HCL or JSON: https://www.terraform.io/docs/providers/nomad/r/job.html and AWS IAM policies inline JSON: https://www.terraform.io/docs/providers/aws/r/iam_policy.html.

Paul's suggestion for us now is that we go with inlining the Config Entry payload but he proposes that we do it as JSON (or at least support JSON too - Nomad can do either). The main reason for this is that in 0.12 people can use the (`jsonencode`)[https://www.terraform.io/docs/configuration/functions/jsonencode.html] built-in function which means they can write pure HCL objects which will still syntax highlight and allow interpolation etc.

```hcl
resource "consul_config_entry" "service-router" {
  kind = "service-router"
  name = "web"
  
  config_json = jsonencode({
    routes = [
      {
        match {
          http {
            path_prefix = "/admin"
          }
        }

        destination {
          service = "admin"
        }
      }
    ]
  })
}
```

I'd be fine with just JSON for now and possibly naming the field explicitly like in example above so that when we do get to update to use 0.12 dynamic types that can be added in a backwards compatible way with just `config` or similar.

What do you think?

remilapeyre:
Hi @banks, thanks for this feedback!

I made the changes to have the configuration in a `config_json` attribute and updated the documentation to show an example with all config entry kinds.

To update `config_json` when reading from Consul I had to marshal, unmarshal and remove some fields at https://github.com/terraform-providers/terraform-provider-consul/pull/127/files#diff-c8616d95befb55df565252daa4b44f84R125-R134, it's not great but I'm not sure it's possible to better as there is no helper function to do this in github.com/hashicorp/consul/api.

An other thing to note, is that if the user does not set fields with default values like `MeshGateway` (https://github.com/terraform-providers/terraform-provider-consul/pull/127/files#diff-88aef67908852df198cd4fd86a154965R78), the next plan will be dirty. I think it could be possible to make them optional by creating two ConfigEntry objects, one with the local configuration and one with thee one read from the API and comparing those objects instead of the JSON strings. I'm not sure it's worth it though.

If the user set a configuration like https://github.com/terraform-providers/terraform-provider-consul/pull/127/files#diff-88aef67908852df198cd4fd86a154965R226-R235, Consul will silently replace the `name` attribute which means we won't be able to read back the config entry from the API and an error will be raised (https://github.com/terraform-providers/terraform-provider-consul/pull/127/files#diff-88aef67908852df198cd4fd86a154965R66). We could hardcode the cases where this happen but it would get cumbersome to maintain and would break as soon as Consul supports namespaces.

Let me know what you think about this implementation, I hope the extensive example in thee documentation will make it easy enough for the user to adapt it to its need despite the less than ideal API :)

banks:
Thanks @remilapeyre!

> I had to marshal, unmarshal

Commented inline but seems like the best option for now.

>if the user does not set fields with default values like MeshGateway the next plan will be dirty

That does seem a bit gross - wont that happen in pretty much every case? I think it's worth doing the deep-equals comparison of structs if it helps avoid that as we expect most fields to be defaulted most of the time especially as we add more. But maybe I'm misunderstanding how often this might trip people up?

> Consul will silently replace the name attribute which means we won't be able to read back the config entry from the API and an error will be raised

Yeah that ones a bit gross but you're right hard coding exceptions is ugly and Consul will change this behaviour "Soon". I guess the key would be documenting the valid value here well, and maybe even calling out the potential error if it's typoed or misconfigured?

What do you think @remilapeyre about adding a "notice" to the docs under the name field that calls out `proxy-defaults` only supports `global` and any other value will cause an error? We'd need to update that notice when namespace support lands but then I think we'll need a bunch of changes to the provider at that point so maybe we can leave a TODO or something we'll remember to grep for there so we don't forget?

remilapeyre:
> That does seem a bit gross - wont that happen in pretty much every case? I think it's worth doing the deep-equals comparison of structs if it helps avoid that as we expect most fields to be defaulted most of the time especially as we add more. But maybe I'm misunderstanding how often this might trip people up?

One thing that worries me with doing deep compares is that `api.DecodeConfigEntry` will happily ignore any mistyped or snake cased attributes, for example this configuration:

``` hcl
resource "consul_config_entry" "proxy_defaults" {
  kind = "proxy-defaults"
  name = "global"

  config_json = jsonencode({
    mesh_gateway = {
      mode = "local"
    }
    Config = {
      local_connect_timeout_ms = 1000
      handshake_timeout_ms     = 10000
    }
  })
}
```

will actually create this config entry:

``` json
    {
        "Kind": "proxy-defaults",
        "Name": "global",
        "Config": {
            "handshake_timeout_ms": 10000,
            "local_connect_timeout_ms": 1000
        },
        "MeshGateway": {},
        "CreateIndex": 16,
        "ModifyIndex": 40
    }
```

and it seems to me that using a deep-equals comparison would not detect any difference here and there is no way to detect that the `mesh_gateway` field has been ignored by `api.DecodeConfigEntry`.

> What do you think @remilapeyre about adding a "notice" to the docs under the name field that calls out proxy-defaults only supports global and any other value will cause an error?

This is a good idea, this way we keep the same behavior as Consul HTTP API but warn the user. I will add this.

remilapeyre:
Hi @banks, if you are OK with it I propose we merge with the current behavior and we improve over it when the support for multiple schemas land in the Terraform SDK.

banks:
Sorry @remilapeyre, I read your last update but never responded.

In your example for why deep equals was bad, is that different from not doing a deep equal check?

If we don't accept a snake_case version in the `api.DecodeConfigEntry` then using snake case will always cause that issue (what is submitted to server won't include the field) whether we do deep equal checking or not wouldn't it? Do we get any better error message now that would help the user?

What would you need exposed in the `api` package to make it clean? If we could expose the snake -> camel normalising in `api.DecodeConfigEntry` for example would that make things work better? That seems totally reasonable to me.

remilapeyre:
Hi @banks, I finally came back to this, I think the UX is better now:
- Optional fields can be omitted from the config and won't generate a dirty plan anymore,
- Extra invalid fields won't be silently discarded anymore but will display an error to the end user

It seems to me that this the best we can do until the support for dynamic fields is ready.

consul/resource_consul_config_entry_test.go
```diff
@@ -0,0 +1,278 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() {},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccConsulConfigEntry_ServiceDefaults,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("consul_config_entry.foo", "name", "foo"),
+					resource.TestCheckResourceAttr("consul_config_entry.foo", "kind", "service-defaults"),
+					resource.TestCheckResourceAttr("consul_config_entry.foo", "config_json", "{\"MeshGateway\":{},\"Protocol\":\"https\"}"),
```
banks: Is this `MeshGateway` key still expected here?

consul/resource_consul_config_entry_test.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func TestAccConsulConfigEntry_Errors(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() {},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccConsulConfigEntry_ProxyDefaultsWrongName,
+				ExpectError: regexp.MustCompile("Provider produced inconsistent result after apply: When applying changes to consul_config_entry.foo, provider \"consul\" produced an unexpected new value for was present, but now absent."),
```
banks: This error message seems like it's missing some words. That might not be something we can fix but assuming this test passes does that mean we are not forming the error message properly somewhere?

consul/resource_consul_config_entry_test.go
```diff
@@ -0,0 +1,278 @@
...
+}
+`
+
+const TestAccConsulConfigEntry_ServiceRouter = `
+resource "consul_config_entry" "web" {
+	name = "web"
+	kind = "service-defaults"
+
+	config_json = jsonencode({
+		MeshGateway = {}
```
banks: I see that we test both with and without this default now so I'm happy it passes both ways but as a nit pick I'd personally leave it out of all but the one test case we actually want to check it works as that reflects typical usage better IMO.

website/docs/r/config_entry.markdown
```diff
@@ -0,0 +1,139 @@
...
+
+## Example Usage
+
+```hcl
+resource "consul_config_entry" "proxy_defaults" {
+  kind = "proxy-defaults"
+  name = "global"
+
+  config_json = jsonencode({
+    MeshGateway = {}
```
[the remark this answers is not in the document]
banks: Can we remove this from the docs now it's not needed, same below etc?

consul/resource_consul_config_entry_test.go
remilapeyre: I think I will need to handle this case separately to give the user a better error message. I will do it in a new PR to not delay this one any longer.

Issue #125, opened 2019-06-27
Allow adding/removing policies for the anonymous token (feature request)


randomswdev:
### Terraform Version
Terraform v0.12.0

### Expected Behavior
If multiple tenants manage the same Consul server, each of them could require adding policies to the anonymous token. While other tokens can be created/destroyed as needed (so every tenant can create its owns), the anonymous token is globally shared and cannot be replicated (Consul manages it in a specific way).
For this reason we would like to have the ability to add/remove policies to that token without impacting the policies added/managed by others.

### Actual Behavior
In the above scenario, currently we can only import the anonymous token (issue #100) in each one of the tenant's terraform files. But as soon as a tenant adds policies, it overwrites the policies set by other tenants.

### References
Issue #100 already discusses the possibility to add logic for updating token's policies. The solution proposed and implemented there does not satisfy our requirement of being able to manage the anonymous token independently across multiple tenants.

### Proposed solution
We would like to define a new resource, for example consul_anonymous_token_policy. This resource will allow to add/remove a single policy from the anonymous token, preserving all the other policies (whoever added them).
If we agree on this design or a possible alternative that would address our need, we can develop a pull request implementing the design.

remilapeyre:
Hi @randomswdev, thanks for opening a new issue. I understand your need, I think this could be added with a new resource:

``` hcl
data "consul_acl_token" "anonymous" {
  id = "00000000-0000-0000-0000-000000000002"
}

resource "consul_acl_policy" "test" {
  name        = "my_policy"
  datacenters = ["dc1"]
  rules       = <<-RULE
    node_prefix "" {
      policy = "read"
    }
    RULE
}

resource "consul_acl_token_policy_attachement" "anonymous" {
  token_id = "${data.consul_acl_token.anonymous.id}"
  policy_id = "${consul_acl_policy.test.id}"
}
```

a bit like `aws_security_group` and `aws_security_group_rule` work in the AWS provider. This would obviously conflicts with the `consul_acl_token.policies` attribute but a warning could be added in the documentation.

I don't think we would be able to distinguish between 

``` hcl
resource "consul_acl_token" "test" {
  description = "my test token"
  policies = []
  local = true
}
```
and
``` hcl
resource "consul_acl_token" "test" {
  description = "my test token"
  local = true
}
```
[...]

PR #130, opened 2019-07-08
Add a new resource that links a token to a policy

randomswdev:
This pull request implements the feature described in issue https://github.com/terraform-providers/terraform-provider-consul/issues/125.
Implemented a new resource to track the link between a token and a policy.

I also checked if it would be possible to prevent consul_acl_token from overwriting the policies if the policies attribute was not included in the resource definition. I think this is not possible without a change to the SDK.
It would be possible to know if an attribute is included or not in the resource definition using the call:
`d.get(parts, getSourceConfig|getSourceExact)`
but the method is private to the schema namespace.

Closes  https://github.com/terraform-providers/terraform-provider-consul/issues/125


consul/resource_consul_acl_token_policy_attachment_test.go
```diff
@@ -0,0 +1,170 @@
...
+	rules = "node \"\" { policy = \"read\" }"
+	datacenters = [ "dc1" ]
+}
+
+resource "consul_acl_token" "test" {
+	description = "test"
+	local = true
+
+	lifecycle {
+		ignore_ttributes = ["policies"]
```
greut: typo?

randomswdev: Yes, the file content was wrong. I just updated it, pushing a new revision.

remilapeyre:
Thanks for opening this PR @randomswdev !

> I also checked if it would be possible to prevent consul_acl_token from overwriting the policies if the policies attribute was not included in the resource definition. I think this is not possible without a change to the SDK.
> It would be possible to know if an attribute is included or not in the resource definition using the call:
> `d.get(parts, getSourceConfig|getSourceExact)`
> but the method is private to the schema namespace.

This still wouldn't be enough: the `local` attribute for example is marked as `ForceNew` and a changing it would  create another token to replace the current one. There is no way to attach the policies to the new token if we don't reapply the `consul_acl_token_policy_attachment` resource too.

randomswdev:
> This still wouldn't be enough: the `local` attribute for example is marked as `ForceNew` and a changing it would create another token to replace the current one. There is no way to attach the policies to the new token if we don't reapply the `consul_acl_token_policy_attachment` resource too.

I just checked this scenario. Changing the `local` attribute, forces a token to be recreated. When a token gets recreated, its id changes . This id is used as the value of the `token_id` attribute of the `consul_acl_token_policy_attachment` resource; because `token_id` is also set as `ForceNew`, the token id's change forces also the attachment to be recreated, thus adding again the policy to the (new) token.

So, I f I understood correctly your scenario, I think the code should work as it is.

remilapeyre:
>>>This id is used as the value of the token_id attribute of the consul_acl_token_policy_attachment resource; because token_id is also set as ForceNew, the token id's change forces also the attachment to be recreated, thus adding again the policy to the (new) token.

Yes but this only works when the `consul_acl_token` and `consul_acl_token_policy_attachment` resources live in the same Terraform configuration, when you have all resources in the same Terraform configuration:

```hcl
resource "consul_acl_policy" "test" {
	name = "test"
	rules = "node \"\" { policy = \"read\" }"
	datacenters = [ "dc1" ]
}
resource "consul_acl_token" "test" {
	description = "test"
	lifecycle {
		ignore_changes = ["policies"]
	}
}
resource "consul_acl_token_policy_attachment" "test" {
    token_id = "${consul_acl_token.test.id}"
    policy = "${consul_acl_policy.test.name}"
}
```
it works well but if you have the token and the policy in separate configuration, you cannot use a reference but have to hardcoded the token as of now :
```hcl
resource "consul_acl_token" "test" {
	description = "test"
	local = true
	lifecycle {
		ignore_changes = ["policies"]
	}
}
```
with in another folder:
```hcl
resource "consul_acl_policy" "test" {
	name = "test"
	rules = "node \"\" { policy = \"read\" }"
	datacenters = [ "dc1" ]
}
resource "consul_acl_token_policy_attachment" "test" {
    token_id = "c597fb2c-6098-474f-9f03-dead07ae2068"
    policy = "${consul_acl_policy.test.name}"
}
```
In this case after changing the `local` parameter of `consul_acl_token.test`, updating the `consul_acl_token_policy_attachment.test` will break as a new UUID will have been generated. I think we could fix this by adding a new `accessor_id` in the `consul_acl_token` resource so the user can choose a permanent UUID (or use https://www.terraform.io/docs/providers/random/r/uuid.html to create one). Does this make sense?

randomswdev:
Added the `accessor_id` to the consul_acl_token resource and updated the relevant documentation.

remilapeyre:
Hi @randomswdev, thanks for adding this, I will make a final review today and will merge this :)

remilapeyre:
Thanks for the help on this @randomswdev 👍 !

PR #133, opened 2019-07-10
Add consul_acl_token and consul_acl_token_secret_id data sources

marimp:
Hi there.
The feature will address the original issue terraform-providers#126 of accessing the secret_id of the consul acl token.
For this purpose a new data source was implemented dataSourceConsulACLToken() .

These are files that implement the data source

    consul/data_source_consul_acl_token.go
    consul/data_source_consul_acl_token_test.go

the required filter parameters and the related outputs

// Datasource predicates
aclTokenAccessorID = "accessor_id"

// Output
aclTokenSecretID    = "secret_id"
aclTokenDescription = "description"
aclTokenPolicies    = "policies"
aclTokenLocal       = "local"

The related documentation file:

    website/docs/assets/d/acl_token.html.markdown

The test related to the new feature was correctly performed:

make testacc
=== RUN   TestAccDataACLToken_basic
--- PASS: TestAccDataACLToken_basic (0.07s)
...

Thanks
Regards

Issue #134, opened 2019-07-30
query_options are not read


marimp:
Hi there,

### Terraform Version
Terraform v0.12.0

### Affected Resource(s)
consul/data_source_consul_services
consul/data_source_consul_nodes
consul/data_source_consul_service

### Expected Behavior
The query_options should be read correctly

### Actual Behavior
The "query_options" attributes are not read because they are wrongly read.
In particular the method `getQueryOpts` in `query_options.go` tries to read them at first level while they are children of attribute query_options

### Steps to Reproduce
You can change the test data_source_consul_service_test.go and change the token attribute to a dummy value. This attribute is not taken into consideration

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-1234

remilapeyre:
Thanks 👏

PR #135, opened 2019-07-30
Fix issue query_options are not read #134

marimp:
This code is to fix defect #134.

I modified the  code for reading `query_options`


remilapeyre:
Thanks @marimp for this PR, the fix looks good to me 👍 

I noticed you closed #133 but it looked good to me. Is this a mistake?

PR #138, opened 2019-08-02
Use consul_keys data source in the index.html example

ccamachofg:
The example to access a key in Consul should be a data source and not a resource


remilapeyre:
Nice catch, the example predates the consul_keys data source :)

PR #141, opened 2019-08-28
Use service_id instead of name when registering health-checks

daktari:
When we register 2 services with same name but different service_id the registrations fails with 500 code error:

2 error(s) occurred:

* consul_service.proxy_http: 1 error(s) occurred:
* consul_service.proxy_http: Failed to register service (dc: ''): Unexpected response code: 500 (rpc error making call: Unknown service 'proxy' for check 'proxy-http-check')

* consul_service.proxy_socks: 1 error(s) occurred:
* consul_service.proxy_socks: Failed to register service (dc: ''): Unexpected response code: 500 (rpc error making call: Unknown service 'proxy' for check 'proxy-socks-check')

This PR fix this issue and makes the correct check service association with service_id instead of service name.

BR

### DIFF
--- a/consul/resource_consul_service.go
+++ b/consul/resource_consul_service.go
@@ -515,7 +515,7 @@ func retrieveService(client *consulapi.Client, name string, ident string, node s
 	return nil, fmt.Errorf("Failed to retrieve service: '%s', services: %v", name, len(services))
 }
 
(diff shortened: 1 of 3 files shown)


remilapeyre:
Thanks @daktari for submitting this change! This was overlooked when we first implementing the health-checks support.

The fix looks good, I just added a test case to make sure we don't trip over this again.

Issue #132, opened 2019-07-10
datasource consul_service does not export attribute "meta"


marimp:
Hi there,

### Affected Data Source
- data_source_consul_service

### Expected Behavior
The attribute "meta" should be exported in the data source "data_source_consul_service":
```
catalogServiceServiceMeta              = "meta"
```
### Actual Behavior
The attribute "meta" was recently added to resource consul_service in issue #122 but it is not implemented in the related datasource.

If you agree I can implement this feature.

Great Thanks
Regards

remilapeyre:
Hi @marimp, this is a nice addition and your commit looks correct. Can you open a new PR with it?

randomswdev:
@remilapeyre I see you have created a pull request for this (and it seems ok to me). Do you have any idea when this change will likely be committe din master? And any idea of when a new version of the provider will be released?

remilapeyre:
Hi @randomswdev, I plan to make a new release in the next days.

PR #144, opened 2019-09-03
Add meta attribute in consul_service data-source

remilapeyre:
Closes https://github.com/terraform-providers/terraform-provider-consul/issues/132

### DIFF
--- a/consul/data_source_consul_service.go
+++ b/consul/data_source_consul_service.go
@@ -115,6 +115,7 @@ func dataSourceConsulService() *schema.Resource {
 						catalogServiceServiceMeta: {
 							Type:     schema.TypeMap,
 							Computed: true,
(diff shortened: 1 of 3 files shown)


consul/data_source_consul_service.go
```diff
@@ -111,6 +112,10 @@ func dataSourceConsulService() *schema.Resource {
 							Computed: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
+						catalogServiceServiceMeta: {
+							Type:     schema.TypeMap,
+							Computed: true,
```
greut: Wouldn't the `Elem` key be required here?

```go
		Elem: &schema.Schema{
			Type: schema.TypeString,
		},
```

remilapeyre: Thanks, I did not know that :)

PR #149, opened 2019-09-07
Remove usage of config pkg

appilon:
In preparation for the standalone tf sdk we are removing usage of the handful of removed packages/functions in the upcoming sdk


appilon:
no new failures in TC

Issue #172, opened 2020-01-03
Adding a License Resource for Consul Enterprise


tryan225:
This is a feature request to add the ability to [update an Consul Enterprise clusters license](https://www.consul.io/api/operator/license.html#updating-the-consul-license) with this provider.

Other use cases could be:
- [resetting the consul license](https://www.consul.io/api/operator/license.html#resetting-the-consul-license)
- [getting the Consul license](https://www.consul.io/api/operator/license.html#getting-the-consul-license)

There is an example of an enterprise-only feature on this list here with `consul_network_segments`:
- GH-46

remilapeyre:
Hi @tryan225, I implemented this in https://github.com/terraform-providers/terraform-provider-consul/pull/173 but I don't have a Consul Enterprise license so I can't test it right now. I will look into how to test this resource now.

PR #173, opened 2020-01-06
Add consul_license resource

remilapeyre:
Closes https://github.com/terraform-providers/terraform-provider-consul/issues/172

### DIFF
--- a/consul/resource_consul_license_test.go
+++ b/consul/resource_consul_license_test.go
@@ -1,28 +1,97 @@
 package consul
 
 import (
+	"os"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 )
 
-func TestAccConsulLicense_basic(t *testing.T) {
+func TestAccConsulLicense_FailOnCommunityEdition(t *testing.T) {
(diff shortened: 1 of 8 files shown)


consul/resource_consul_license_test.go
```diff
@@ -0,0 +1,28 @@
+package consul
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccConsulLicense_basic(t *testing.T) {
```
pearkes: Is it possible to add some kind of conditional test that runs against the enterprise version?

remilapeyre: Do you think it's OK if the test look for a `test_license.hcli` in the working dir and run if it finds one?

pearkes: Seems reasonable with documentation and logging. Another option is an API query to /v1/agent/self that maybe interrogates the version string or similar?

remilapeyre: I did both, some resources like the `consul_network_segment` or the namespace support can be run without actually having a license when using the `hashicorp/consul-enterprise` Docker image.

PR #182, opened 2020-02-17
Add import support to consul_acl_role

greut:
the ACL role should be importable just as a policy is.


remilapeyre:
Hi @greut! Thanks for the PR!

greut:
@remilapeyre thanks for putting the final touch to it. :+1:

PR #184, opened 2020-02-25
Fix error message in consul_acl_token datasource

greut:
fix error message


remilapeyre:
I need to hide my copy-pasting better ;)

Thanks for the correction @greut !

PR #190, opened 2020-04-21
Fix typo in documentation: s/gpg_key/pgp_key/

greut:
The documentation mentions `gpg_key` when the attribute is `pgp_key`


remilapeyre:
Hi @greut, thanks for the patch :)

PR #198, opened 2020-06-15
Fix typo in consul_service documentation

ricardbejarano:
"Register an health-check" to "Register a health-check"


remilapeyre:
Thanks, I'm not a native speaker and it's the kind of subtleties I tend to get wrong.

PR #203, opened 2020-06-29
Add filtering support to consul_service and consul_service_health data sources

remilapeyre:


consul/data_source_consul_service_health.go
```diff
@@ -16,84 +16,81 @@ func dataSourceConsulServiceHealth() *schema.Resource {
...
-				// ForceNew: true,
-				Default: true,
+				Default:  true,
 			},
-			"wait_for": &schema.Schema{
+			"wait_for": {
+				Optional: true,
+				Type:     schema.TypeString,
+			},
+			"filter": {
```
mkeeler, line 51: Should the filter have `ForceNew: true` or is that not necessary to make terraform make an api request with the new filter?

mkeeler: After a little more thought I am fairly certain its not needed and TF will do the right thing as-is. Getting some confirmation on that point would be great though.

remilapeyre: Good question, for datasources `ForceNew` is not needed as a request they only implement `Read` they will always make an API call no matter whether their arguments changed. `ForceNew` is only meaningful for resources that implements `Update`, for resources that don't implement `Update` all arguments must have `ForceNew: true`.

remilapeyre:
Thanks for the review @mkeeler !

consul/data_source_consul_service_health.go
mkeeler: 👍

PR #210, opened 2020-07-23
Allow creating tokens with pregenerated secret_ids

brknstrngz:
Hi,

I am aware of how contentious this is due to sensitive data ending up in the state file, so I am just testing the water.

In a scenario where token IDs are programatically generated and distributed to the Consul, Nomad and Vault nodes over an out of band channel, they can then be subsequently provisioned in Consul. Perhaps this can also be placed under a boldface warning sign as we have for consul_acl_token_secret_id, suggesting storing the state remotely?

Thanks!

### DIFF
--- a/consul/resource_consul_acl_token.go
+++ b/consul/resource_consul_acl_token.go
@@ -20,12 +20,20 @@ func resourceConsulACLToken() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"secret_id": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Computed:    true,
+				Optional:    true,
+				Sensitive:   true,
+				Description: "The token secret id.",
+			},
 			"accessor_id": {
 				Type:        schema.TypeString,
 				ForceNew:    true,
 				Computed:    true,
 				Optional:    true,
-				Description: "The token id.",
+				Description: "The token accessor id.",
 			},
 			"description": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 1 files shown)


remilapeyre:
Hi @brknstrngz, you are correct that this would end up saving the token in the state file which is why we did not offer this possibility in the first place.

At the moment the ACL bootstrap more manual that I would like it to be, but there is work going on in Consul to use JWT to simplify this part: https://github.com/hashicorp/consul/issues/7972. Once this is ready we should be able to use this to securely add nodes to the cluster.

As of today, I don't have a better solution, if you need tokens with a specific `secret_id` you could create them manually as you already do and import them in Terraform using `terraform import consul_acl_token.my_token ...` and then migrate to the JWT method to introduce new nodes when it's ready.

brknstrngz:
> As of today, I don't have a better solution, if you need tokens with a specific `secret_id` you could create them manually as you already do and import them in Terraform using `terraform import consul_acl_token.my_token ...` and then migrate to the JWT method to introduce new nodes when it's ready.

Thank you @remilapeyre, thank you for the swift reply!

My understanding of this is that there needs to be an existing Consul cluster that the provider configuration points to, where tokens to be imported reside. Is that correct? If so, I may be in a chicken and egg situation - at the point in time where my token secret ids are generated the Consul cluster has not yet been provisioned.

My workflow is as follows:

1. Generate the master token for the Consul servers and the agent tokens and policies for all Consul nodes (servers and clients), Nomad servers and clients and Vault
2. Have Terraform provision cloud VMs
3. Use a configuration management tool (Salt, in this case) to place the agent token IDs in the Consul, Nomad and Vault configuration files on all nodes
4. Wait for Consul server quorum
5. Using the tokens and policies generated in step 1, programatically generate a new Terraform state containing the configuration for this provider and all consul_acl_policy and consul_acl_token entries, then apply that state.

Seth-Karlo:
Hi, just to revive this a little bit, I've put together almost an identical PR myself, with some documentation explaining the risks of storing the tokens in the state file. With appropriate warnings, would this be acceptable to merge to match existing Consul API functionality?

PR #214, opened 2020-08-12
Fix datacenter handling in consul_intention

### LINKED ISSUE #213: Consul Intentions do not support datacenter argument
### Terraform Version
Currently running 0.12.18

### Affected Resource(s)
- consul_intention

### Terraform Configuration Files
```hcl
resource "consul_intention" "allow_foo" {
  source_name        = "my-service"
  datacenter            = "dc1"
  destination_name = "my-other-service"
  action                    = "allow"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
N/A

### Expected Behavior
The [docs for consul_intention](https://www.terraform.io/docs/providers/consul/r/intention.html) state that datacenter should be an optional argument for the resource, but it d 
### Actual Behavior
What actually happened?

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

### Important Factoids
N/A

### References
Using `consul_node` as an example it looks like the argument is defined [here](https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_node.go#L24-L29) with fallback to the provider defined datacenter [here](https://github.com/terraform-providers/terraform-provider-consul/blob/master/consul/resource_consul_node.go#L60-L67). It doesn't look like `consul_intention` has this although the docs suggest it should. 

I'd be happy to put a PR together to address this, but wanted to get the issue up first in case I'm missing something important.

tjrivera:
This patch should resolve #213, so that the `consul_intention` resource accepts datacenter as an argument, which is already reflected in existing documentation.

### DIFF
--- a/consul/resource_consul_config_entry_test.go
+++ b/consul/resource_consul_config_entry_test.go
@@ -114,7 +114,7 @@ func TestAccConsulConfigEntry_basic(t *testing.T) {
 			{
 				Config: testAccConsulConfigEntry_TerminatingGateway(extraConf),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("consul_config_entry.terminating_gateway", "name", "foo"),
+					resource.TestCheckResourceAttr("consul_config_entry.terminating_gateway", "name", "foo-egress"),
 					resource.TestCheckResourceAttr("consul_config_entry.terminating_gateway", "kind", "terminating-gateway"),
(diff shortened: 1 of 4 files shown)


remilapeyre:
Thank you very much for reporting this issue and working on the patch @tjrivera !

PR #220, opened 2020-08-18
Allow specifying TLS Certificates via PEM strings

lawliet89:
- Instead of paths only

```
Running tool: /usr/bin/go test -timeout 30s -run ^TestResourceProvider_ConfigureTLSPem$

2020/08/18 17:20:21 [INFO] Initializing Consul client
2020/08/18 17:20:21 [INFO] Consul Client configured with address: 'demo.consul.io:80', scheme: 'https', datacenter: 'nyc3', insecure_https: 'false'
PASS
ok  	github.com/terraform-providers/terraform-provider-consul/consul	0.029s

```

### DIFF
--- a/consul/config.go
+++ b/consul/config.go
@@ -16,8 +16,11 @@ type Config struct {
 	HttpAuth      string `mapstructure:"http_auth"`
 	Token         string `mapstructure:"token"`
 	CAFile        string `mapstructure:"ca_file"`
+	CAPem         string `mapstructure:"ca_pem"`
 	CertFile      string `mapstructure:"cert_file"`
+	CertPEM       string `mapstructure:"cert_pem"`
 	KeyFile       string `mapstructure:"key_file"`
+	KeyPEM        string `mapstructure:"key_pem"`
 	CAPath        string `mapstructure:"ca_path"`
 	InsecureHttps bool   `mapstructure:"insecure_https"`
 	Namespace     string `mapstructure:"namespace"`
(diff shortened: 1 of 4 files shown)


remilapeyre:
Hi @lawliet89, this looks mainly good to me, I just have one question: why did you remove the `ConflictsWith` you had set in the first place? It looks like they would actually be useful here.

lawliet89:
@remilapeyre 

Consider the following configuration

```hcl
provider "consul" {
  address = "consul.example.com"
  scheme  = "https"
  ca_pem  = file("/ca/ca-chain.pem")
}

resource "consul_keys" "test" {
  key {
    path  = "test/test"
    value = "foobar"
  }
}
```

Terraform will error with 

```
$ terraform plan

Error: "ca_pem": conflicts with ca_file



Error: "ca_file": conflicts with ca_pem
```

This is most likely because of usage of `schema.EnvDefaultFunc` like `schema.EnvDefaultFunc("CONSUL_CA_FILE", "")` which gives `""` by default. Terraform does not think `""` is `null`.

remilapeyre:
@lawliet89, thanks for explaining. Can you try with the changes I pushed? I tested it and it should work but I would like to be sure.

lawliet89:
Thanks @remilapeyre! TIL that the function can take `nil` without issues. It works for me.

remilapeyre:
> Side question: I notice that the environment variable supported here are different from what the Consul CLI supports. Is there a reason for that? Should we also support what Consul CLI supports?

I'm not sure why there is some additional environment variables in the Terraform provider, they were added before I worked on the project. All environment variables supported by the Consul CLI because in https://github.com/hashicorp/terraform-provider-consul/blob/8f65aef6090652e665082ab4e0000923cf93ba17/consul/config.go#L27-L88 we call `consulapi.DefaultConfig()` that reads the environment variables to set some attributes. It did happen that the defaults in the Terraform provider clashed with the environment variables and they were ignored in some cases but we should have fixed those. Do you have one in particular in mind?

remilapeyre:
Hi @lawliet89 thanks for the work you did to get this issue fixed. As I said in my previous message the Terraform provider should behave like the CLI in regard to the environment variables. If you think you found a bug, please don't hesitate to report it in a new issue :)

lawliet89:
Thanks @remilapeyre Your explanation makes sense.

PR #236, opened 2020-10-31
Use docker mirror

mdeggies:
CHANGELOG: no impact

Dockerhub is going to rate limit unauthenticated pulls on November 1st. IPs from CI machines (with the exception of github acitons) will be near their limit all the time. We're moving projects to use a public un-rate-limited Mirror of these images instead. Let me know if you have any q's, otherwise feel free to merge when you can!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,14 @@ install:
 - bash scripts/gogetcookie.sh
 
 script:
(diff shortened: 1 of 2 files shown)


remilapeyre:
Thanks for taking care of this @mdeggies ! If we start using another docker image in our CI will we need to ask someone to add it to the mirror or will it look in the Dockerhub and cache it automatically?

PR #241, opened 2021-02-08
Clarify the wording in the docs for consul_key_prefix import

ryndaniels:
I'm not entirely sure if this is the exact correct wording, but it looked like the existing description for the import needed a bit of re-working. Happy to iterate on this to make sure it makes sense to everyone and accurately reflects the import behavior! 🙂


preetapan (approved):

mkeeler (approved):

remilapeyre (approved):
Thanks @ryndaniels, this wording is indeed better than the previous one :)

PR #242, opened 2021-02-22
Add namespaced token support to the consul_acl_token_secret_id data source

mkeeler:
This PR enabled the tf provider to retrieve the token secret for ACL tokens scoped to a namespace.


remilapeyre (approved):
Thanks for taking care of this @mkeeler !

PR #268, opened 2021-05-26
Migrate tests from Travis to Actions

mdeggies:
Travis-ci.org is being deprecated at the end of this week, so we're moving things to GitHub actions. Tests run using go 1.14.x, 1.15.x, and 1.16.x on the ubuntu executor.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,92 @@
...
+  test:
+    env:
+      GO111MODULE: on
+      GOFLAGS: -mod=vendor
+      CONSUL_HTTP_TOKEN: master-token
+      CONSUL_HTTP_ADDR: http://localhost:8500
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [1.14.x, 1.15.x, 1.16.x]
```
alvin-huang, line 18: do we need all 3 versions? I only saw 1.16 in travis.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,92 @@
+on:
```
alvin-huang, line 1: can you rename this file to `.yml` 😄

alvin-huang (approved):
LGTM

PR #277, opened 2021-07-28
docs: Add deprecation note for consul_intentions resource

blake:
Note that as of Consul 1.9 intentions should be managed using `consul_config_entry` instead of the `consul_intentions` resource.


remilapeyre (approved):
Thanks!

PR #288, opened 2021-09-19
Add note regarding perpetual diffs in consul_config_entry documentation

remilapeyre:
Closes https://github.com/hashicorp/terraform-provider-consul/issues/281

### DIFF
--- a/website/docs/r/config_entry.markdown
+++ b/website/docs/r/config_entry.markdown
@@ -12,6 +12,11 @@ The [Configuration Entry](https://www.consul.io/docs/agent/config_entries.html)
 resource can be used to provide cluster-wide defaults for various aspects of
 Consul.
 
+~> **NOTE:** Because the schema in a `consul_config_entry` resource can very
+widely between the various configuration entry kinds it is necessary to define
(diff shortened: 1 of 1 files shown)


website/docs/r/config_entry.markdown
```diff
@@ -12,6 +12,11 @@ The [Configuration Entry](https://www.consul.io/docs/agent/config_entries.html)
 resource can be used to provide cluster-wide defaults for various aspects of
 Consul.
 
+~> **NOTE:** Because the schema in a `consul_config_entry` resource can very
+widely between the various configuration entry kinds it is necessary to define
+explicitely every attributes or Terraform will always report a diff on the
+resource.
```
blake, line 18:
```suggestion
~> **NOTE:** Because the schema in a `consul_config_entry` resource can very
widely between the various configuration entry kinds, it is necessary to explicitly
define every attribute to avoid Terraform reporting a diff on the resource.
```

blake (approved):
I just had a minor suggestion for changing the phrasing, and correcting a minor spelling error. Otherwise, LGTM.

PR #297, opened 2022-02-23
feat: lazy creation of the Consul client

greut:
The goal of this PR is to be able to use dynamic data in the configuration of the provider, such as the _address_ field coming from a data source. To achieve this, the actual creation of the Consul client must be delayed as much as possible.

We might be losing some error catching (it'd need a much bigger PR), hence there is a attempt to decode the configuration and fails early whenever possible.


greut:
I've got the same errors running the tests on the trunk.

remilapeyre:
Hi @greut, the change looks great. There is quite a few tests that broke with the latest version of Consul. I fixed the tests in https://github.com/hashicorp/terraform-provider-consul/pull/292 and will review this PR tomorrow.

greut:
@remilapeyre cool. After all, it seems our tests are working without that change.

greut:
Interesting.

remilapeyre:
Hi @greut, I'm coming back to this feature as I'm preparing the new release and I'm now a bit confused about it. It seems to me like this should be handled by Terraform core and that there is nothing special to do on the provider side (though I may have botched the merge from master during my initial pass over the code so feel free to revert it if I overwrote something important).

Basically when having a dependency from one provider to another like this:

```terraform
terraform {
  required_providers {
    random = {
      source = "hashicorp/random"
      version = "3.1.2"
    }

    consul = {
      source = "hashicorp/consul"
      version = "2.14.0"
    }
  }
}

resource "random_string" "random" {
  length = 16
}

provider "consul" {
  address = random_string.random.result
}

resource "consul_acl_policy" "agent" {
  name  = "agent"
  rules = <<-RULE
    node_prefix "" {
      policy = "read"
    }
  RULE
}
```

Terraform core should notice the dependency and instantiate the providers properly. The output of `terraform graph | dot -Tsvg > graph.svg` shows that the dependency is handled as I would expect:

graph

The arrow from `provider["registry.terraform.io/hashicorp/consul"]` to `random_string.random` shows that the provider should not be configured before the `random_string.random` resource has been created.

When running `terraform apply` I get:

```
random_string.random: Creating...
random_string.random: Creation complete after 0s [id=VG6syqjtK2vvNpX]]
consul_acl_policy.agent: Creating...
╷
│ Error: error creating ACL policy: Put "http://VG6syqjtK2vvNpX]/v1/acl/policy": dial tcp: address VG6syqjtK2vvNpX]:80: unexpected ']' in address
│ 
│   with consul_acl_policy.agent,
│   on example.tf line 23, in resource "consul_acl_policy" "agent":
│   23: resource "consul_acl_policy" "agent" {
│ 
╵
```

showing what I would expect.

Is this not your experience? Am I missing something?

greut:
> Is this not your experience?

Pretty much. If i'm using this branch locally, all the CI runs are made with the vanilla provider without any issues. It seems not needed. Thanks a lot for the time you've put into that.

Cheers,

PR #309, opened 2022-06-17
Add support for Consul Peering

remilapeyre:


.github/workflows/test.yaml
```diff
@@ -29,8 +29,8 @@ jobs:
       run: docker run -v $PWD:/src docker.mirror.hashicorp.services/bflad/tfproviderlint -S006 -S022 -S023 ./...
     - name: Run OSS acceptance tests
       run: |
-        curl -LO https://releases.hashicorp.com/consul/1.11.4/consul_1.11.4_linux_amd64.zip
-        sudo unzip consul_1.11.4_linux_amd64.zip consul -d /usr/local/bin
+        curl -LO https://releases.hashicorp.com/consul/1.13.0-alpha1/consul_1.13.0-alpha1_linux_amd64.zip
```
mkeeler, line 32: These should probably be changed to the 1.13.1 release.

consul/data_source_consul_peering.go
```diff
@@ -0,0 +1,103 @@
...
+	return &schema.Resource{
+		Read: dataSourceConsulPeeringRead,
+		Schema: map[string]*schema.Schema{
+			// In
+			"peer_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"datacenter": {
```
mkeeler, line 20: We should probably get rid of this parameter. Consul will not forward the HTTP/RPC requests to read a peering to other datacenters so this would have no effect.

consul/data_source_consul_peering.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+			"peer_server_addresses": {
+				Type:     schema.TypeList,
+				Computed: true,
+
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
```
mkeeler, line 71: We also added ImportedServiceCount and ExportedServiceCount fields to the Peering type in the api package. Should those also be included in the data source?

vendor/github.com/google/go-cmp/cmp/compare.go
```diff
@@ -36,7 +36,6 @@ import (
 	"strings"
```
mkeeler, line 36: On a meta note, should we get rid of the vendor directory? Most other hashicorp repos have.

consul/data_source_consul_peering_test.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+}
+
+const testAccDataConsulPeeringMissing = `
+data "consul_peering" "basic" {
+  peer_name = "hello"
+}
+`
+
+const testAccDataConsulPeeringBasic = `
```
mkeeler, line 46: This test likely wont run properly against 1.13. We specifically prevent users from peering a partition/cluster with itself. I guess the path forward for testing purposes is to use a couple provider aliases and multiple consul dev agents.

consul/data_source_consul_peerings.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func dataSourceConsulPeerings() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceConsulPeeringsRead,
+		Schema: map[string]*schema.Schema{
+			"datacenter": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
mkeeler, line 17: Same as previous comment: datacenter isn't currently applicable to peerings.

Also the same two extra fields are also present in the response

consul/data_source_consul_peerings_test.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+	})
+}
+
+const testAccDataConsulPeeringsNone = `
+data "consul_peerings" "basic" {}
+`
+
+const testAccDataConsulPeeringsBasic = `
+resource "consul_peering_token" "basic" {
```
mkeeler, line 56: ditto: the test will need multiple test agents and provider aliases to work.

consul/resource_consul_peering.go
```diff
@@ -0,0 +1,171 @@
...
+				ForceNew:    true,
+				Sensitive:   true,
+				Description: "The peering token fetched from the peer cluster.",
+			},
+			"datacenter": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Specifies the datacenter where the peering is established. Defaults to the agent's datacenter when not specified.",
+			},
```
mkeeler, line 49: ditto: Datacenter field is unused with the final 1.13 GA builds.

consul/resource_consul_peering.go
```diff
@@ -0,0 +1,171 @@
...
+				Computed: true,
+			},
+			"peer_server_addresses": {
+				Type:     schema.TypeList,
+				Computed: true,
+
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
```
mkeeler, line 99: Ditto about the two additional fields that the read will output?

For the resource we should suppress the diffs for those values as they are meant as informational to the user and not something the user can set/control.

consul/resource_consul_peering_token.go
```diff
@@ -0,0 +1,103 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"datacenter": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Specifies the datacenter where the peering token is generated. Defaults to the agent's datacenter when not specified.",
+			},
```
mkeeler, line 48: ditto: no need for the datacenter field.

consul/resource_consul_peering_token.go
```diff
@@ -0,0 +1,103 @@
...
+The functionality described here is available only in Consul version 1.13.0 and later.
+`,
+
+		Create: resourceConsulPeeringTokenCreate,
+		Read: func(*schema.ResourceData, interface{}) error {
+			return nil
+		},
+		Delete: func(*schema.ResourceData, interface{}) error {
+			return nil
+		},
```
mkeeler, line 29: hmm, so peering token create will behind the scenes create a peering that can be read or deleted. The read and delete functions should be able to reuse what you created for the consul_peering resource. Without support for deletions, you could not use the tf provider to manage peerings without leaking the accepting side  of the peering relationship.

mkeeler (commented):
Thanks for putting this PR up. There have been a few changes since the alpha which will necessitate some changes but overall, almost everythingin the PR looks correct to me.

consul/resource_consul_peering.go
```diff
@@ -0,0 +1,171 @@
...
+		Create: resourceConsulPeeringCreate,
+		Read:   resourceConsulPeeringRead,
+		Delete: resourceConsulPeeringDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			// In
+			"peer_name": {
```
mkeeler, line 31: This resource needs a `partition` field as well.

consul/resource_consul_peering_token.go
remilapeyre: Hi @mkeeler, I'm not sure what else could do here, there's support for deleting a peering connection at https://github.com/remilapeyre/terraform-provider-consul/blob/peering/consul/resource_consul_peering.go#L163-L173 but the Consul API does not provide a way to invalid a Peering token that has not yet been used to established a connection.

Support for doing this with the Consul API would be needed if we want to implement a meaningful Delete() for this resource.

mkeeler: It does work but the peering APIs are a bit asymmetrical:

```
➜  curl -X POST localhost:8500/v1/peering/token -d '{"PeerName": "foo"}'
{
    "PeeringToken": "eyJDQSI6bnVsbCwiU2VydmVyQWRkcmVzc2VzIjpbIjEyNy4wLjAuMTo4NTAyIl0sIlNlcnZlck5hbWUiOiJzZXJ2ZXIuZGMxLmNvbnN1bCIsIlBlZXJJRCI6IjRmZmJkMzljLWZkYzYtOThkOC1kMWE3LWU3MjM3ZTEyMjllOSIsIkVzdGFibGlzaG1lbnRTZWNyZXQiOiJmYTE0M2Y1MS1kNDU1LThiYWQtNWM5Yi0yMDFhZmY2NTk4YzQifQ=="
}

~
➜  curl localhost:8500/v1/peerings
[
    {
        "ID": "4ffbd39c-fdc6-98d8-d1a7-e7237e1229e9",
        "Name": "foo",
        "Partition": "default",
        "State": "PENDING",
        "ImportedServiceCount": 0,
        "ExportedServiceCount": 0,
        "CreateIndex": 18,
        "ModifyIndex": 18
    }
]

~
➜  curl -X DELETE localhost:8500/v1/peering/foo

~
➜  curl localhost:8500/v1/peerings
[]
```

Basically issuing a request to either `POST /v1/peering/token` or `POST /v1/peering/establish` will create a peering which can be deleted with a `DELETE /v1/peering/<peer name>`.

mkeeler: Similarly you can still read an individual peering that has a token generated for it but the stream is not yet established:

```
➜  curl localhost:8500/v1/peering/foo
{
    "ID": "947b29ca-8dcf-1d61-1da8-0defeda6e0f0",
    "Name": "foo",
    "Partition": "default",
    "State": "PENDING",
    "ImportedServiceCount": 0,
    "ExportedServiceCount": 0,
    "CreateIndex": 27,
    "ModifyIndex": 27
}
```

remilapeyre: Hi @mkeeler, all should be good now.

mkeeler (approved):

PR #315, opened 2022-08-03
Support partitioned tokens in ACL secret id data source

mkeeler:
## Description

This PR adds support for partitions to the `consul_acl_token_secret_id` resource.

Tokens can now be created with the `consul_acl_token` resource in non-default partitions and then have their secret ids retrieved with the data source.


remilapeyre (approved):
Thanks @mkeeler!

PR #318, opened 2022-08-28
Add consul_config_entry datasource

remilapeyre:
I'm not updating the docs in this pull request as I will switch the
whole provider to use terraform-plugin-docs in
https://github.com/hashicorp/terraform-provider-consul/pull/309. I will
update the documentation after merging this PR.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/317


consul/data_source_consul_config_entry.go
```diff
@@ -0,0 +1,74 @@
...
+		return fmt.Errorf("failed to read config entry %s/%s: %w", kind, name, err)
+	}
+
+	// Config Entries are too complex to write as maps for now so we save their JSON representation
+	data, err := configEntryToMap(configEntry)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s", kind, name))
```
dhiaayachi, line 68: Wouldn't id collide with other config entries with the same kind/name in other partitions/namespaces?

remilapeyre: `id` in Terraform resources or datasources is often unique but does not need to, when objects don't have a specific ID it can happen that they have the same of another. For example Consul has this at https://github.com/hashicorp/terraform-provider-consul/blob/master/consul/resource_consul_keys.go#L188-L191 where all `consul_keys` resources will have the same `id` no matter the keys being set and another occurence of this in the Vault provider is at https://github.com/hashicorp/terraform-provider-vault/blob/main/vault/resource_generic_endpoint.go#L111 where multiple `vault_generic_endpoint` resources can have the same `id` despite being in different namespace.

Here because config entries do not have IDs and the user won't use it it does not matter what placeholder we use.

dhiaayachi (approved):
🚀

PR #319, opened 2022-08-28
Add import support to consul_config_entry

remilapeyre:
Related to https://github.com/hashicorp/terraform-provider-consul/issues/316


consul/resource_consul_config_entry.go
```diff
@@ -16,6 +16,38 @@ func resourceConsulConfigEntry() *schema.Resource {
...
+				case 4:
+					kind = parts[0]
+					name = parts[1]
+					partition = parts[2]
+					namespace = parts[3]
+				default:
+					return nil, fmt.Errorf(`expected path of the form "<kind>/<name>" or "<kind>/<name>/<partition>/<namespace>"`)
+				}
+
+				d.SetId(fmt.Sprintf("%s-%s", kind, name))
```
dhiaayachi, line 36: Shouldn't partition and namespace be added to the ID to ensure uniqueness?

templates/resources/config_entry.md
```diff
@@ -280,3 +280,15 @@ The following attributes are exported:
...
 
 * `config_json` - A map of configuration values.
+
+
+## Import
+
+`consul_config_entry` can be imported using the syntax `<kind>/<name>` if the
+config entry is in the default partition and default namespace, or
+`<kind>/<name>/<partition>/<namespace>` for config entries in a non-default
+partition or namespace:
```
dhiaayachi, line 290: What about providing `foo/bar/default/default` is it equivalent to `foo/bar`?

dhiaayachi (commented):
👋 @remilapeyre 
I had few questions about that one, specifically around partition and namespace handling.

templates/resources/config_entry.md
remilapeyre: Yes it Is equivalent. Looking at it now it seems to me like `<partition>/<namespace>/<kind>/<name>` might make more sense here.

consul/resource_consul_config_entry.go
remilapeyre: The ID does not need to be unique. For config entries there is no ID that exist so we can use any placeholder there. We have other resources that not have a unique ID either, there is some examples at: https://github.com/hashicorp/terraform-provider-consul/pull/318#discussion_r1046452455

templates/resources/config_entry.md
dhiaayachi: I agree, that would be more inline with consul annotations around namespace and partition.

remilapeyre: Done!

PR #321, opened 2022-08-28
Mark token attribute in resources and datasources as deprecated

remilapeyre:
Having the token attribute in the resources causes issues: when a user
updates it the previous value is used when calling the `Read()` method.

This is expected based on the protocol Terraform Core uses and there is
no way to have the correct behavior here. To fix this I mark the "token"
attribute as deprecated so that users only set the one in the "provider"
block.

I also mark all "datacenter", "namespace" and "partition" attribute as
ForceNew to avoid the same issue with those.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/298


dhiaayachi (approved):
LGTM! 🚢

PR #323, opened 2022-09-09
Allow importing consul nodes

marceloboeira:
It seems the import function is overall "necessary" since creating a node that already exists ends up updating it / binding it to the terraform resource. But I could also not find strong reasons NOT to allow import since it facilitates a more standard and "terraform compliant" workflow when migrating things from consul configs to terraform.

I have tested both using `$name-$address` and simply `$name`, but it seems unnecessary, at first, to use the address since the only logical isolations possible would be namespace/datacenter.


remilapeyre:
Hi @marceloboeira, as you mentioned the `consul_node` resource actually does not need import support due to its behaviour: just creating a new resource will work and replace the current node configuration just like importing it would.

I understand that this may be confusing as it may not be what some users may be expecting so it makes sense to make it also possible to imported.

Thanks for this PR!

PR #326, opened 2022-10-20
Update resource consul_service documentation to include partition

davidfleming:
Added partition to docs.

https://github.com/hashicorp/terraform-provider-consul/blob/edeffd2ae33747e0a65d5e4a9cb7398f16b4ecbb/consul/resource_consul_service.go#L209


remilapeyre (approved):
Thanks for your submission @davidfleming !

PR #328, opened 2022-12-07
Update for Consul 1.14

remilapeyre:


docs/resources/service.md
```diff
@@ -110,8 +110,6 @@ of the `name` attribute.
 
 * `namespace` - (Optional, Enterprise Only) The namespace to create the service within.
 
-* `partition` - (Optional, Enterprise Only) The partition the service is associated with.
-
```
dhiaayachi, line 114: I'm not sure I understand why this is removed, isn't service partition a valid option?

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/hashicorp/terraform-provider-consul
 
 require (
-	github.com/hashicorp/consul/api v1.14.0
+	github.com/hashicorp/consul/api v1.16.0
```
dhiaayachi, line 4: I think API version release with 1.14.2 is 1.18.0 https://github.com/hashicorp/consul/blob/0ba7a4018304e81c370b662aa435fee45d287076/go.mod#L34

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.17.0
+
+NEW FEATURES:
+
+* The `consul_config_entry` now supports specifying the name of the peered cluster to export the service to.
+
```
dhiaayachi, line 6: I don't see a code change in the PR related to that, is that done in a prior PR?

dhiaayachi (commented):
👋 @remilapeyre 
Thank you for working on this!
I left few comments/questions.

go.mod
remilapeyre: Thanks.

docs/resources/service.md
remilapeyre: This was a mistake, thanks.

CHANGELOG.md
remilapeyre: The handling of config entries in the provider is a bit special: we call `consulapi.DecodeConfigEntry()` at https://github.com/hashicorp/terraform-provider-consul/blob/master/consul/resource_consul_config_entry.go#L150. This way we don't need to make changes to the provider when there is changes to them, we just update the module used by the provider and the new version of `consulapi.DecodeConfigEntry()` is called.

dhiaayachi (approved):
🚀

PR #341, opened 2023-05-01
Support complex configuration in consul_certificate_authority

remilapeyre:
When configuration Consul to use Vault as the certificate authority provider for Consul Connect the configuration can have more than strings so we need to use jsonencode for Terraform to accept the configuration.

Also remove the copyright notices in the examples.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/324


sorenisanerd:
I haven't reviewed the code, but I'm running into the issue that this is aiming to address. Specifically, I want to use AppRole between Consul and Vault, and there's currently no way to trick Terraform into passing the `Params` struct that Consul expects.

dhiaayachi (approved):
@remilapeyre A comment about the general approach here, and please disregard if you think that's not achievable. Do you think it's possible instead of introducing `config_json` we can accept the new format in `config` and at the read time, check if it's json or not by attempting an unmarchall?  As I see both fields have the same type schema and decode to `map[string]interface{}`

PR #342, opened 2023-05-01
Add support to log with a JWT auth method

remilapeyre:
This is an alternative of giving a Consul token in the provider configuration or as en environment variable and can be use in Terraform Cloud runs by setting the `use_terraform_cloud_workload_identity` parameter.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/334


dhiaayachi (approved):
LGTM!

examples/data-sources/consul_peering/data-source.tf
```diff
@@ -1,6 +1,3 @@
-# Copyright (c) HashiCorp, Inc.
-# SPDX-License-Identifier: MPL-2.0
```
roncodingenthusiast, line 2: any reason why we are deleting these? we have been adding copyright headers to all our files

remilapeyre: This is the example files that are used to generate the documentation, for example at https://registry.terraform.io/providers/hashicorp/consul/latest/docs/data-sources/peering.

None of the providers have this notice in their example and the scaffolding repository exclude them explicitely at https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.copywrite.hcl

PR #345, opened 2023-05-07
Validate that the policies argument in consul_acl_role contains UUIDs

remilapeyre:
Related to https://github.com/hashicorp/terraform-provider-consul/issues/344


dhiaayachi (approved):
LGTM!

PR #355, opened 2023-09-25
Net 5676 - Service Defaults, Service Splitter, Service Resolver, Service Router and Service Intentions Resource  In Terraform Consul Provider

absolutelightning:
Net 5676 - Service Defaults, Service Splitter, Service Resolver, Service Router and Service Intentions Resource In Terraform Consul Provider

Tested by CI and by `terraform init` and `terraform apply`


consul/resource_service_defaults_config_entry.go
```diff
@@ -0,0 +1,557 @@
...
+		},
+		"balance_outbound_connections": {
+			Type:     schema.TypeString,
+			Optional: true,
+		},
+	},
+}
+
+var serviceDefaultsConfigEntrySchema = map[string]*schema.Schema{
+	"kind": {
```
Ganeshrockz, line 118: Is `kind` still required given that we are moving away from the `config_entry` resource? Wouldn't it always be `service-defaults`?

consul/resource_service_defaults_config_entry.go
```diff
@@ -0,0 +1,557 @@
...
+		},
+		Schema: serviceDefaultsConfigEntrySchema,
+	}
+}
+
+func fixQOptsForServiceDefaultsConfigEntry(name, kind string, qOpts *consulapi.QueryOptions) {
+	// exported-services config entries are weird in that their name correspond
+	// to the partition they are created in, see
+	// https://www.consul.io/docs/connect/config-entries/exported-services#configuration-parameters
+	if kind == "exported-services" && name != "default" {
```
Ganeshrockz, line 374: Should this be removed?

consul/resource_service_defaults_config_entry.go
```diff
@@ -0,0 +1,557 @@
...
+			return fmt.Errorf(`failed to read config entry after setting it.
+This may happen when some attributes have an unexpected value.
+Read the documentation at https://www.consul.io/docs/agent/config-entries/%s.html
+to see what values are expected`, configEntry.GetKind())
+		}
+		return fmt.Errorf("failed to read config entry: %v", err)
+	}
+
+	d.SetId(fmt.Sprintf("%s-%s", kind, name))
+	return resourceConsulServiceDefaultsConfigEntryRead(d, meta)
```
Ganeshrockz, line 517: We seem to do a `Get` on the same resource for setting the `Id` attribute. Any reason to do it again here?

consul/resource_service_defaults_config_entry.go
```diff
@@ -0,0 +1,557 @@
...
+	configName := d.Get("name").(string)
+
+	if _, err := configEntries.Delete(configKind, configName, wOpts); err != nil {
+		return fmt.Errorf("failed to delete '%s' config entry: %v", configName, err)
+	}
+	d.SetId("")
+	return nil
+}
+
+func makeServiceDefaultsConfigEntry(kind, name string, configMap map[string]interface{}, namespace, partition string) (consulapi.ConfigEntry, error) {
```
Ganeshrockz, line 545: I don't think `kind` needs to be passed as a param to this method. You could move the string `service-defaults` as a constant and reference it whenever needed

consul/resource_service_defaults_config_entry_ce_test.go
```diff
@@ -0,0 +1,310 @@
+// Copyright (c) HashiCorp, Inc.
```
Ganeshrockz, line 1: What is the difference between `ce_test` and `ee_test`?

Ganeshrockz (commented):
Looks good in general. Will review the test files later on.

consul/resource_service_defaults_config_entry.go
absolutelightning: It is computed. User can not enter this. Required for linke `474` to set kind.

consul/resource_service_defaults_config_entry_ce_test.go
absolutelightning: namespace partition is added.

consul/resource_service_defaults_config_entry.go
absolutelightning: ok

consul/resource_service_defaults_config_entry.go
absolutelightning: First we call the update API then we do a get to confirm that resource is created with that id. If get gives an error that means update failed.

absolutelightning: This similar thing is done for config_entry resource as well.

consul/resource_service_defaults_config_entry.go
absolutelightning: yes

consul/resource_service_defaults_config_entry_ce_test.go
absolutelightning: also it runs on enterprise binary

consul/helper.go
```diff
@@ -0,0 +1,94 @@
...
+		if token == "tls" {
+			keyToReturn += strings.ToUpper(token)
+		} else {
+			keyToReturn += strings.ToTitle(token)
+		}
+	}
+	return keyToReturn
+}
+
+func FormatKeys(config interface{}, formatFunc func(string) string) (interface{}, error) {
```
Ganeshrockz, line 42: Should this be a public function?

consul/helper.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package consul
```
Ganeshrockz, line 4: Do you see a need to have a separate set of test cases for this file? Especially the `FormatKeys` method

Ganeshrockz (approved):
LGTM

consul/helper.go
absolutelightning: no. will add that.

consul/helper.go
absolutelightning: changed.

consul/resource_consul_service_splitter_config_entry.go
```diff
@@ -0,0 +1,255 @@
...
+
+const KindServiceSplitter = "service-splitter"
+
+var serviceSplitterConfigEntrySchema = map[string]*schema.Schema{
+	"name": {
+		Type:     schema.TypeString,
+		Required: true,
+		ForceNew: true,
+	},
+	"kind": {
```
remilapeyre, line 22: I think we don't need the kind here because it should always be service-splitter

consul/resource_consul_service_splitter_config_entry.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return &schema.Resource{
+		Create: resourceConsulServiceSplitterConfigEntryUpdate,
+		Update: resourceConsulServiceSplitterConfigEntryUpdate,
+		Read:   resourceConsulServiceSplitterConfigEntryRead,
+		Delete: resourceConsulServiceSplitterConfigEntryDelete,
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				parts := strings.Split(d.Id(), "/")
+				var kind, name, partition, namespace string
```
remilapeyre, line 129: We also don't need the kind during import because it will always be `service-splitter`.

remilapeyre (changes requested):

consul/resource_consul_service_splitter_config_entry.go
absolutelightning: This is computed. So user will not be able to enter this.

consul/resource_consul_service_splitter_config_entry.go
absolutelightning: This is computed. So user will not be able to enter this.

absolutelightning: This field is computed. There is not interaction of this field with user. Also I am having trouble removing it because 
```				sw.set("kind", kind)
				sw.set("name", name)
				sw.set("partition", partition)
				sw.set("namespace", namespace)
```
state writer is setting in resouces data itself.

absolutelightning: What is the issue in keeping it this way?

consul/resource_consul_service_splitter_config_entry.go
```diff
@@ -0,0 +1,255 @@
...
+
+func resourceConsulServiceSplitterConfigEntryRead(d *schema.ResourceData, meta interface{}) error {
+	client, qOpts, _ := getClient(d, meta)
+	configEntries := client.ConfigEntries()
+	configKind := d.Get("kind").(string)
+	configName := d.Get("name").(string)
+
+	fixQOptsForConfigEntry(configName, configKind, qOpts)
+
+	_, _, err := configEntries.Get(configKind, configName, qOpts)
```
remilapeyre, line 226: You need to use the result from `configEntries.Get()` to update the Terraform state with d.Set() or stateWriter.set()

remilapeyre (changes requested):

consul/resource_consul_config_entry_service_defaults.go
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	getTransparentProxy := func(transparentProxy map[string]interface{}) *consulapi.TransparentProxyConfig {
+		return &consulapi.TransparentProxyConfig{
+			OutboundListenerPort: transparentProxy["outbound_listener_port"].(int),
+			DialedDirectly:       transparentProxy["dialed_directly"].(bool),
+		}
+	}
+
+	getEnvoyExtension := func(envoyExtensionMap map[string]interface{}) *consulapi.EnvoyExtension {
```
remilapeyre, line 455:
```suggestion
	getEnvoyExtension := func(envoyExtensionMap map[string]interface{}) consulapi.EnvoyExtension {
```

You can return a `consulapi.EnvoyExtension` here because it is what you need later.

remilapeyre (changes requested):

PR #356, opened 2023-09-25
added cas support for consul_keys resource

aahel:
added cas support for consul_keys terraform


consul/resource_consul_keys_test.go
```diff
@@ -42,6 +42,20 @@ func TestAccConsulKeys_basic(t *testing.T) {
...
 
+func TestAccConsulKeys_Cas(t *testing.T) {
+	providers, _ := startTestServer(t)
+
+	resource.Test(t, resource.TestCase{
+		Providers: providers,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccConsulKeysConfig_Cas,
+				Check:  resource.TestCheckResourceAttr("consul_keys.app", "key.2637474718.cas", "0"),
```
absolutelightning, line 53: Should we check for path and value as well here along with cas.

consul/resource_consul_keys.go
```diff
@@ -154,7 +157,17 @@ func resourceConsulKeysCreateUpdate(d *schema.ResourceData, meta interface{}) er
...
 
 			flags := sub["flags"].(int)
-
+			cas := sub["cas"].(int)
+			if cas != -1 {
+				written, err := keyClient.Cas(path, value, flags, cas)
+				if err != nil {
+					return err
+				}
+				if written {
```
absolutelightning, line 166: what if written is false

aahel: removed it. since if written is false still the key should not get removed

absolutelightning (approved):

PR #363, opened 2023-10-09
Support both UUIDs and names in consul_acl_role.policies

remilapeyre:
The `consul_acl_role` attribute `policies` only support UUIDs like:

```terraform
resource "consul_acl_policy" "read-policy" {
  name        = "read-policy"
  rules       = "node \"\" { policy = \"read\" }"
}

resource "consul_acl_role" "read" {
  name        = "foo"

  policies = [
    consul_acl_policy.read-policy.id
  ]
}
```

This differs from the Consul API payload where the `Policies` attribute is actually a list of objects, not a list of strings:

```json
{
  "Name": "example-role",
  "Policies": [
    {
      "ID": "783beef3-783f-f41f-7422-7087dc272765"
    }
  ]
}
```

This makes it possible to set the policies using either the ID, or the name:

```json
{
  "Name": "example-role",
  "Policies": [
    {
      "ID": "783beef3-783f-f41f-7422-7087dc272765"
    },
    {
      "Name: "Test"
    }
  ]
}
```

This is not supported by the `consul_acl_role` resource.

Not being able to add a policy using its name makes things harder for users and hanging its schema to make `policies` a list of objects would be best, but it would also break the backward compatibility of the provider which I always try not to.

This patch makes it possible possible to use either and tries first to fetch the corresponding policy using the ID, then using the name.

`resourceConsulACLRoleRead()` is also updated to set the correct value in Terraform so that we don't have a perpetual diff.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/344.


consul/resource_consul_acl_role.go
```diff
@@ -298,5 +320,38 @@ func getRole(d *schema.ResourceData, meta interface{}) *consulapi.ACLRole {
...
-	return role
+	return role, nil
+}
+
+// getPolicyByIdOrName looks for a policy in Consul first by ID, then by name if
+// it found nothing. It also returns a boolean indicating whether the identifier
+// given is the ID or the name
+func getPolicyByIdOrName(identifier string, client *consulapi.Client, qOpts *consulapi.QueryOptions) (*consulapi.ACLPolicy, bool, error) {
+	policy, _, initialErr := client.ACL().PolicyRead(identifier, qOpts)
+	if policy != nil && initialErr == nil {
```
dhiaayachi, line 331: you can rely on both being nil here to detect a not-found use case. I think it would simplify the logic.

dhiaayachi (approved):
I left a comment that could lead to simplified code, but it LGTM!

PR #367, opened 2023-10-11
Implement the consul_config_entry_service_intentions resource

absolutelightning:


consul/resource_consul_config_entry_service_intentions.go
```diff
@@ -0,0 +1,548 @@
...
+)
+
+type serviceIntentions struct{}
+
+func (s *serviceIntentions) GetKind() string {
+	return consulapi.ServiceIntentions
+}
+
+func (s *serviceIntentions) GetDescription() string {
+	return "The `consul_service_intentions_config_entry` resource configures a [service intentions](https://developer.hashicorp.com/consul/docs/connect/config-entries/service-intentions) that are configurations for controlling access between services in the service mesh. A single service intentions configuration entry specifies one destination service and one or more L4 traffic sources, L7 traffic sources, or combination of traffic sources."
```
remilapeyre, line 20:
```suggestion
	return "The `consul_service_intentions_config_entry` resource configures [service intentions](https://developer.hashicorp.com/consul/docs/connect/config-entries/service-intentions) that are configurations for controlling access between services in the service mesh. A single service intentions configuration entry specifies one destination service and one or more L4 traffic sources, L7 traffic sources, or combination of traffic sources."
```

consul/resource_consul_config_entry_service_intentions.go
```diff
@@ -0,0 +1,548 @@
...
+						Default:     "consul",
+						Description: "Specifies the type of destination service that the configuration entry applies to.",
+						Optional:    true,
+					},
+					"description": {
+						Type:        schema.TypeString,
+						Description: "Specifies a description of the intention.",
+						Optional:    true,
+					},
+					"legacy_id": {
```
remilapeyre, line 233:
```suggestion
					"legacy_id": {
```

We should not expose any legacy attribute through Terraform: they might be remove in the future and nobody is relying on them when using Terraform for now so let's not add them for now and if someone needs them we can always add them later.

remilapeyre (changes requested):

PR #375, opened 2023-10-27
Consul keys data

NightOwl998:
Adding a new attribute for the provider `error_on_missing_key` , if set to true the provider will send an error when reading data sources for missing keys, and if set to false (by default it is set to false), return an empty value when reading a non existing key.


remilapeyre (approved):
Thank you @NightOwl998 !

PR #379, opened 2023-11-20
Don't panic on missing role in consul_acl_role_policy_attachment

remilapeyre:
`resourceConsulACLRolePolicyAttachmentCreate()` and `resourceConsulACLRolePolicyAttachmentDelete()`
were not checking properly if the given role ID was not found in Consul.

Some functions in the Consul SDK return an error when the object is missing,
other return a `nil` object.

In the case of `RoleRead()` I was checking for the wrong thing. I checked the
rest of the ACL resources and as far as I know it, this was the only
resource impacted.

I also update the `consul_acl_policy` data source to use the new
`PolicyReadByName()` function.

Closes https://github.com/hashicorp/terraform-provider-consul/issues/378

### DIFF
--- a/consul/data_source_consul_acl_policy.go
+++ b/consul/data_source_consul_acl_policy.go
@@ -6,7 +6,6 @@ package consul
 import (
 	"fmt"
 
-	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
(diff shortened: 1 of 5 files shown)


consul/data_source_consul_acl_policy.go
```diff
@@ -53,25 +52,13 @@ func dataSourceConsulACLPolicyRead(d *schema.ResourceData, meta interface{}) err
...
-	}
-
-	policy, _, err := client.ACL().PolicyRead(policyEntry.ID, qOpts)
+	policy, _, err := client.ACL().PolicyReadByName(name, qOpts)
 	if err != nil {
 		return fmt.Errorf("could not read policy '%s': %v", name, err)
 	}
+	if policy == nil {
+		return fmt.Errorf("could not find policy %q", name)
+	}
```
dhiaayachi, line 61: I just fixed the same bug in Consul CLI 😆 
https://github.com/hashicorp/consul/pull/19679

dhiaayachi (approved):
LGTM!

PR #399, opened 2024-02-27
Add V2 exported services config support

skpratt:
uses the small api module change merged here: https://github.com/hashicorp/consul/pull/20737


skpratt:
@absolutelightning looks like I cannot add you as a reviewer

consul/data_source_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+	}
+}
+
+func dataSourceConsulV2ExportedServicesRead(d *schema.ResourceData, meta interface{}) error {
+	client, qOpts, _ := getClient(d, meta)
+	name := d.Get("name").(string)
+	kind := d.Get("kind").(string)
+	gvk := &GVK{ //&api.GVK{
+		Group:   "multicluster",
```
absolutelightning, line 87: This can be replaced by `pbmulticluster.GroupName`
and Version can be replaced by `pbmulticluster.Version`

consul/data_source_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,133 @@
...
+	if err = protojson.Unmarshal(respData, data); err != nil {
+		return fmt.Errorf("failed to unmarshal to proto message: %v", err)
+	}
+	respID, err := json.Marshal(resp["id"])
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal response id: %v", err)
+	}
+	id := &pbresource.ID{}
+	if err = protojson.Unmarshal(respID, id); err != nil {
+		return fmt.Errorf("Failed to unmarshal to proto message: %v", err)
```
absolutelightning, line 109:
```suggestion
		return fmt.Errorf("failed to unmarshal to proto message: %v", err)
```

consul/data_source_consul_config_entry_v2_exported_services_test.go
```diff
@@ -0,0 +1,116 @@
...
+					resource.TestCheckResourceAttr("data.consul_config_entry_v2_exported_services.read", "peer_consumers.0", "peer1"),
+				),
+			},
+		},
+	})
+}
+
+const testAccDataSourcePartitionExportedServicesV2ConfigBasic = `
+resource "consul_config_entry_v2_exported_services" "test" {
+	name = "test"
```
absolutelightning, line 60: Is this indentation intended?

consul/resource_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,205 @@
...
+	if err = protojson.Unmarshal(respData, data); err != nil {
+		return fmt.Errorf("failed to unmarshal to proto message: %v", err)
+	}
+	respID, err := json.Marshal(resp["id"])
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal response id: %v", err)
+	}
+	id := &pbresource.ID{}
+	if err = protojson.Unmarshal(respID, id); err != nil {
+		return fmt.Errorf("Failed to unmarshal to proto message: %v", err)
```
absolutelightning, line 167:
```suggestion
		return fmt.Errorf("failed to unmarshal to proto message: %v", err)
```

consul/resource_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,205 @@
...
+	sw.set("sameness_group_consumers", samenessgroups)
+	return sw.error()
+}
+
+func resourceConsulV2ExportedServicesDelete(d *schema.ResourceData, meta interface{}) error {
+	client, qOpts, _ := getClient(d, meta)
+	gvk := &GVK{
+		Group:   "multicluster",
+		Version: "v2",
+		Kind:    "ExportedServices",
```
absolutelightning, line 201:
```suggestion
		Kind:     pbmulticluster.ExportedServicesKind,
```

consul/resource_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	samenessConsumers := d.Get("sameness_group_consumers").([]interface{})
+	for _, sg := range samenessConsumers {
+		consumers = append(consumers, map[string]any{"sameness_group": sg})
+	}
+	data := map[string]any{"consumers": consumers}
+	services := d.Get("services").([]interface{})
+	if len(services) > 0 {
+		data["services"] = services
+	}
```
absolutelightning, line 121: Should we return if `data["services"]` is blank?

consul/resource_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,205 @@
...
+	client, _, wOpts := getClient(d, meta)
+	name := d.Get("name").(string)
+	kind := d.Get("kind").(string)
+	gvk := &GVK{
+		Group:   "multicluster",
+		Version: "v2",
+		Kind:    kind,
+	}
+	var consumers []map[string]any
+	peerConsumers := d.Get("peer_consumers").([]interface{})
```
absolutelightning, line 105: All the Get calls with `d.Get` will fail of user has not given the key in schema. From the documentation 
```
// Get returns the data for the given key, or nil if the key doesn't exist
// in the schema.
```
Since these are optional keys we need to add nil checks everywhere.

consul/v2_resource_provider_helper.go
```diff
@@ -0,0 +1,63 @@
...
+	Data       map[string]any    `json:"data"`
+	Owner      *pbresource.ID    `json:"owner,omitempty"`
+	ID         *pbresource.ID    `json:"id"`
+	Version    string            `json:"version"`
+	Generation string            `json:"generation"`
+	Status     map[string]any    `json:"status"`
+}
+
+func v2MulticlusterRead(client *api.Client, gvk *GVK, resourceName string, q *api.QueryOptions) (map[string]interface{}, error) {
+	endpoint := strings.ToLower(fmt.Sprintf("/api/%s/%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind, resourceName))
```
absolutelightning, line 37: Are we converting the `resourceName`to lower case as well?
Is this intended?

consul/data_source_consul_config_entry_v2_exported_services.go
```diff
@@ -0,0 +1,133 @@
...
+	gvk := &GVK{ //&api.GVK{
+		Group:   "multicluster",
+		Version: "v2",
+		Kind:    kind,
+	}
+	resp, err := v2MulticlusterRead(client, gvk, name, qOpts)
+	if err != nil || resp == nil {
+		return fmt.Errorf("exported services config not found: %s", name)
+	}
+	respData, err := json.Marshal(resp["data"])
```
dhiaayachi, line 95: I think if the "data" field don't exist (for some reason) this will return an empty `pbmulticluster.ExportedServices` struct, is that the behaviour we want? 🤔
Same for "ID"

consul/data_source_consul_config_entry_v2_exported_services_test.go
skpratt: it seems to be the convention for the other data_source_consul_*_test.go cases

consul/resource_consul_config_entry_v2_exported_services.go
skpratt: no, because NamespaceExportedServices and PartitionExportedServices do not explicitly set services

consul/resource_consul_config_entry_v2_exported_services.go
skpratt: refactored to in-line the loop - this pattern is used with optional keys everywhere

absolutelightning (approved):

consul/resource_consul_acl_binding_rule_test.go
```diff
@@ -74,7 +74,7 @@ func TestAccConsulACLBindingRule_basic(t *testing.T) {
 			},
 			{
 				Config:      testResourceACLBindingRuleConfig_wrongType,
-				ExpectError: regexp.MustCompile(`Invalid Binding Rule: unknown BindType "foobar"`),
+				ExpectError: regexp.MustCompile(`unknown BindType "foobar"`),
```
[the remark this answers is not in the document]
skpratt: The error message changed slightly with the version updates in .github/workflows/test.yaml, and the test started to fail

PR #421, opened 2024-11-19
Removes pre-release cluster peering warning from documentation

nickwales:
The cluster peering feature is no longer in pre-release.

### DIFF
--- a/docs/resources/peering.md
+++ b/docs/resources/peering.md
@@ -14,8 +14,6 @@ description: |-
 
 The `cluster_peering` resource can be used to establish the peering after a peering token has been generated.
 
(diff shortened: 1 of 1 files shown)


nathancoleman:
@nickwales the docs are generated, so I believe you'll need to remove the source of this warning here

https://github.com/hashicorp/terraform-provider-consul/blob/21e154216f672b98cd15bf1ec91b2ee6e2e253b0/consul/resource_consul_peering.go#L21-L22

and then regenerate the docs just to make sure everything is clean. Not any docs that I could find for this process, but you'll need to do the same thing the CI does here:

https://github.com/hashicorp/terraform-provider-consul/blob/45a6e54c98f1ef608416dfecbb26f8c514b717f1/.github/workflows/test.yaml#L47-L50

PR #435, opened 2025-05-14
Add support for configuring weights on Consul services

### LINKED ISSUE #310: How to add weights to a service
I am creating a service to be used with HAProxy  `server-template`, something like this:

```
resource "consul_service" "test" {
  name       = "test"
  service_id = "test-2"
  node       = consul_node.test.name
  port       = 443
  weights = {
    passing = 100,
    warning = 1
  }
}
```

But I get:
```
╷
│ Error: Unsupported argument
│
│   on lb_test.tf line 6, in resource "consul_service" "test":
│    6:   weights = {
│
│ An argument named "weights" is not expected here.╵
```

From here, consul seems to support weights: https://github.com/hashicorp/consul/pull/4468/files but how to use them from terraform?

t-davies:
Resolves #310.

- Adds support for configuring [service weights](https://developer.hashicorp.com/consul/docs/reference/service#weights).


consul/resource_consul_service.go
```diff
@@ -594,6 +601,11 @@ func getCatalogRegistration(d *schema.ResourceData, meta interface{}) (*consulap
 	}
 	registration.Service.Meta = serviceMeta
 
+	if serviceWeights, ok := d.GetOk("weights"); ok {
+		registration.Service.Weights.Passing = serviceWeights.(map[string]interface{})["passing"].(int)
+		registration.Service.Weights.Warning = serviceWeights.(map[string]interface{})["warning"].(int)
```
t-davies, line 606: Need to check if these are set, just because we have `weights` doesn't mean these exist.

consul/resource_consul_service.go
```diff
@@ -375,6 +382,11 @@ func resourceConsulServiceRead(d *schema.ResourceData, meta interface{}) error {
 	sw.set("namespace", service.Namespace)
 	sw.set("partition", service.Partition)
 
+	sw.set("weights", map[string]int{
+		"passing": service.ServiceWeights.Passing,
+		"warning": service.ServiceWeights.Warning,
+	})
```
t-davies, line 388: We need to treat "" in state and "1" from Consul as equivalent, otherwise we'll always get a diff in the scenario where `weights` are not set.

Vikramarjuna (approved):
LGTM!

consul/resource_consul_service_test.go
```diff
@@ -355,6 +357,41 @@ func TestAccConsulService_datacenter(t *testing.T) {
...
+	providers, _ := startTestServer(t)
+
+	resource.Test(t, resource.TestCase{
+		Providers: providers,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccConsulServiceConfigBasicWeightsOptional,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckNoResourceAttr("consul_service.example", "weights.passing"),
+					resource.TestCheckResourceAttr("consul_service.example", "weights.warning", "5"),
```
Vikramarjuna, line 388: This UT seems to be failing

t-davies: Yep, thanks - not sure how I missed that! We didn't set default values and therefore the weights weren't being set at all in the case where we only provided one of them, should be fixed now.

Vikramarjuna (approved):

